PR #25, opened 2017-09-27
http extension configured

dmytro-landiak:


src/main/java/org/thingsboard/gateway/extensions/http/HttpController.java
```diff
@@ -22,29 +22,28 @@
...
     @Autowired
-    private SigfoxService service;
+    private HttpService service;
 
     private ObjectMapper mapper = new ObjectMapper();
 
 
-    @RequestMapping(value = "/{deviceTypeId}", method = RequestMethod.POST)
-    public void handleRequest(@PathVariable String deviceTypeId,
+    @RequestMapping(value = "/{baseUrl}/{converterId}", method = RequestMethod.POST)
```
ashvayka: Please add two similar methods, or two mappings 
@RequestMapping(value = "/sigfox/{deviceTypeId}", method = RequestMethod.POST)
@RequestMapping(value = "/uplink/{converterId}", method = RequestMethod.POST)

ashvayka:
One minor comment, otherwise LGTM.

src/main/java/org/thingsboard/gateway/extensions/http/HttpController.java
dmytro-landiak: okey, will be fixed.

PR #34, opened 2017-12-14
tb-gateway refactoring: file persistent store and netty-mqtt client

mp-loki:


src/main/java/org/thingsboard/gateway/service/MqttGatewayService.java
```diff
@@ -328,25 +333,12 @@ private void checkDeviceConnected(String deviceName) {
...
-            });
-            return new MqttDeliveryFuture(token);
-        } catch (MqttException e) {
-            onFailure.accept(e);
-            return new MqttDeliveryFuture(e);
+        if (mqttMessageQueue.size() < connection.getMaxQueueSize()) {
+            MqttMessageWrapper messageWrapper = new MqttMessageWrapper(topic, msg, onSuccess, onFailure, System.currentTimeMillis());
+            mqttMessageQueue.add(messageWrapper);
         }
+        return null;
```
ashvayka: is this ok?

mp-loki: well, I was thinking to change the interface return type to void, as soon as we are decoupling recieving and sending of an mqtt message

src/main/java/org/thingsboard/gateway/service/MqttMessageSender.java
```diff
@@ -0,0 +1,92 @@
...
+        this.persistence = persistence;
+        this.tbClient = tbClient;
+        this.mqttSenderService = mqttSenderService;
+        this.messageRepository = messageRepository;
+    }
+
+    @Override
+    public void run() {
+        while (true) {
+            List<PersistentMqttMessage> storedMessages = null;
```
ashvayka: In case of any exception this thread will die and everything will stop working. Please add try catch and logging.

src/main/java/org/thingsboard/gateway/service/MqttMessageServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+@Slf4j
+public class MqttMessageServiceImpl implements MqttMessageService {
+
+    @Autowired
+    private PersistentMqttMessageRepository messageRepository;
+
+    @Autowired
+    private TbPersistenceConfiguration persistence;
+
+    private Map<UUID, Consumer<PersistentMqttMessage>> successCallbacks;
```
ashvayka: ConcurrentMap interface may be useful.

src/main/java/org/thingsboard/gateway/service/MqttMessageServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        if (future != null) {
+            future.complete(null);
+        }
+    }
+
+    @Override
+    public void resolveFutureFailed(UUID id) {
+        MqttDeliveryFuture future = futures.remove(id);
+        if (future != null) {
+            future.complete(null);
```
ashvayka: Maybe you should have Boolean (true/false)? How to check that delivery failed?

src/main/java/org/thingsboard/gateway/service/MqttMessageWrapper.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private final String topic;
+    private final String deviceId;
+    private final int messageId;
+    private final byte[] payload;
+    private final long timestamp;
+
+
+    @Override
+    public int compareTo(MqttMessageWrapper that) {
```
ashvayka: what if timestamp matches? maybe compare device ids?

pom.xml
```diff
@@ -96,17 +105,42 @@
...
+            <artifactId>spring-integration-jpa</artifactId>
+            <version>${spring.version}</version>
+        </dependency>
         <dependency>
             <groupId>org.bouncycastle</groupId>
             <artifactId>bcpkix-jdk15on</artifactId>
             <version>${bouncycastle.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.hsqldb</groupId>
```
ashvayka: Please make sure that the HSLDB is compacted periodically.

src/main/java/org/thingsboard/gateway/service/MqttMessageSender.java
mp-loki: Done

src/main/java/org/thingsboard/gateway/service/MqttMessageServiceImpl.java
mp-loki: Done, although I don't use anything else than get() and put()

src/main/java/org/thingsboard/gateway/service/MqttMessageWrapper.java
mp-loki: Obsolete class - removed. PersistentMqttMessage is now used instead, it has UUID as ID

src/main/java/org/thingsboard/gateway/service/MqttDeliveryFuture.java
```diff
@@ -15,66 +15,11 @@
...
-    @Override
-    public Void get() throws InterruptedException, ExecutionException {
-        return get(Optional.empty());
-    }
-
-    @Override
-    public Void get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        return get(Optional.of(unit.toMillis(timeout)));
-    }
+public class MqttDeliveryFuture extends CompletableFuture<Boolean> {
```
ashvayka: :+1:

PR #55, opened 2018-04-06
Timestamp and format mapping for MQTT

AlexanderJP91:
Added the possibility to map timestamp and timestamp format using MQTT extension

### DIFF
--- a/src/main/java/org/thingsboard/gateway/extensions/mqtt/client/conf/mapping/MqttJsonConverter.java
+++ b/src/main/java/org/thingsboard/gateway/extensions/mqtt/client/conf/mapping/MqttJsonConverter.java
@@ -35,6 +35,8 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import java.text.ParseException;
+
 /**
  * Created by ashvayka on 23.01.17.
  */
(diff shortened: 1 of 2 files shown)


mp-loki:
@AlexanderJP91 Thank you for contributing!

sfergusonpoc:
Does this mean that I can now send client-side timestamps in the form of mosquitto_pub -h localhost -p 1883 -t "sensors" -m '{"ts":1451649600512,"values":{"serialNumber":"SN-001", "model":"T1000", "temperature":36.6}}'  ? I've never been able to post the client-side time stamp successfully.

AlexanderJP91:
@sfergusonpoc Sure you can now send client side timestamps for both timeseries and attributes. You need to configure the MQTT JSON parsing inside the gateway as follows:
`
{
	"topicFilter": "Specified topic",
	"converter": {
		"type": "json",
		"filterExpression": "",
		"deviceNameJsonExpression": "JSON filter expression",
		"timeseries": [
		{
			"type": "type of timeseries variable",
			"key": "JSON path to the key variable",
			"value": "JSON path to the timeseries variable",
			"ts": "JSON path to the ts variable",
			"tsFormat": "JSON path to the tsFormat variable"
		}]
	}
}
`

NOTE: there could be a bug when "deviceNameJsonExpression" is defined, since the JSON root path changes, so I suggest you to first try implementing it without specifying the deviceNameJsonExpression.

sfergusonpoc:
Thanks for the help, was able to get this to work finally.

PR #111, opened 2018-08-09
Fixed get configuration request payload

lonwern:
The right request payload should be `{"sharedKeys":"configuration"}` instead of `{"shared":"configuration"}`.

TB will return all attributes if the request payload was wrong, and once the response payload bigger than  8192 bytes, Gateway will be crashed.

```
[nioEventLoopGroup-2-1] WARN  i.n.channel.DefaultChannelPipeline - An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.NullPointerException: null
	at nl.jk5.mqtt.MqttPingHandler.channelRead(MqttPingHandler.java:34)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362)
...
```

image


felix-ma:
Thank you very much. I also found this problem.

Issue #138, opened 2018-11-08
MQTT extension configuration update did not delete old configuration.


felix-ma:
When I update the configuration, every update will regenerate a client.
Causes a request to have a duplicate return, and even the previous configuration can cause the program to go wrong.
I found that there is a way of destroying in source code.
![image](https://user-images.githubusercontent.com/30289895/48206550-a27ed780-e3a9-11e8-9a50-7d0ef56592b7.png)
In theory, the old configuration stopped before the configuration took effect. In fact, when I repeatedly modified the configuration, I received many feedback messages.

When I modified the configuration, the log display configuration came into effect.
```
22:59:34.690 [pool-2-thread-2] INFO  o.t.g.e.m.c.DefaultMqttClientService - [Tenant] Initializing MQTT client service!
22:59:34.691 [pool-2-thread-2] DEBUG o.t.g.e.m.client.MqttBrokerMonitor - [localhost:2883] MQTT broker connection attempt!
22:59:34.696 [MQTT Call: fe5b6379-3144-4523-8b64-6cc53444362e] INFO  o.t.g.e.m.client.MqttBrokerMonitor - [localhost:2883] MQTT broker connection established!
22:59:34.697 [pool-2-thread-2] INFO  o.t.g.s.gateway.MqttGatewayService - Subscription added: AttributesUpdateSubscription(deviceNameFilter=.*, listener=org.thingsboard.gateway.extensions.mqtt.client.MqttBrokerMonitor@245d206f)
22:59:34.697 [pool-2-thread-2] INFO  o.t.g.s.gateway.MqttGatewayService - Subscription added: RpcCommandSubscription(deviceNameFilter=.*, listener=org.thingsboard.gateway.extensions.mqtt.client.MqttBrokerMonitor@245d206f)
22:59:35.689 [pool-3-thread-1] DEBUG o.t.g.service.MqttMessageSender - Sending message [{deviceId='null', payload={"mqttExtensionStatus":"Updated"}, timestamp=0, topic='v1/devices/me/telemetry'id=9e4e7ac0-8458-430d-adf4-bfff55a54ab4, messageId=15}]
22:59:35.690 [pool-3-thread-1] DEBUG o.t.g.service.MqttMessageSender - Sending message [{deviceId='null', payload={"mqttExtensionError":""}, timestamp=0, topic='v1/devices/me/telemetry'id=0195a803-f3dd-4824-a372-40c10856331c, messageId=16}]
[...]

PR #139, opened 2018-11-08
Fix MQTT extension configuration update did not delete old configuration. #138

felix-ma:
#138


mp-loki:
Thank you for contributing!

PR #189, opened 2019-06-05
Fixed MODBUS first connection failure without attempting connection

felix-ma:
I found that when the Modbus service disconnects from the start gateway, it only tries to connect once, and when the Modbus service starts, it does not automatically connect.
I modified the connection logic following mqtt's retry method


zbeacon:
Hi, @felix-ma 

Improved functionality.

You can try using a new gateway for your purposes instead of a Java gateway.

PR #190, opened 2019-06-12
 add modbus Order of bytes BIG_ENDIAN_BYTE_SWAP

guodaxia103:
1、 BIG_ENDIAN_BYTE_SWAP  ： 2 byte and 2 byte exchange
2、" .asText() " method is null ,This is BUG  ， modify  .toString();


zbeacon:
Hi, @guodaxia103 

Improved functionality.

Please consider adding your functionality to new version. If you are not familiar with Python I can guide you through the required changes or port your changes. Which one do you prefer?

PR #192, opened 2019-06-13
Update MqttGatewayService.java

Zimuge:
the params filed in payload is an json array, so can't be parsed as string, we need to deserialization params to String by using toString() not asText();


guodaxia103:
我已经提交过此bug 了， 你也在搞 RPC 哈~ 一起交流以下

tohidprogram:
public abstract String asText()
Method that will return a valid String representation of the container value, if the node is a value node (method isValueNode() returns true), otherwise empty String its better to use:
```
rpcCommand.setParams(data.get("params").toString());
```

zbeacon:
Hi, @Zimuge 

Improved functionality.

You can try using a new gateway for your purposes instead of a Java gateway.

PR #213, opened 2019-09-10
fix null reference error. 

lincoln310:
When use remote config, the field "filterExpression" generated from ui should use some valid string or an empty space, if not ,this will be null. But if user does not want to use this, the empty space will be forgot to set. So we should check it here.


zbeacon:
Hi, @lincoln310 
Improved functionality.

The new gateway doesn't uses filter expressions. You can try to use it instead of a Java gateway.

PR #227, opened 2019-11-06
Added value parsing and posting depending on tag

Sliosas:
I was not able to post values from this setup:

https://gist.github.com/Sliosas/73022f887a591dcfb8d0c74dde2931ac

to thingsboard-server. The problem is that i have a modbus device that returns two values in one response (modbus tcp response example):

b'\x00\x01\x00\x00\x00\x07\x01\x04\x04\x00\x34\x00\x44'

however thingsboard uses only the first value for both (Windspeed and Direction) timeseries tags and then fails to create a correct json object for the telemetry key.

Am i using a bad mobus.json configuration, is it not possible to return multiple values in one call, or is it a real issue?

### DIFF
--- a/thingsboard_gateway/connectors/modbus/bytes_modbus_uplink_converter.py
+++ b/thingsboard_gateway/connectors/modbus/bytes_modbus_uplink_converter.py
@@ -28,7 +28,7 @@ class BytesModbusUplinkConverter(ModbusConverter):
         for config_data in data:
             if self.__result.get(config_data) is None:
                 self.__result[config_data] = []
-            for tag in data[config_data]:
+            for tag_indx, tag in enumerate(data[config_data]):
                 data_sent = data[config_data][tag]["data_sent"]
                 input_data = data[config_data][tag]["input_data"]
                 log.debug("Called convert function from %s with args", self.__class__.__name__)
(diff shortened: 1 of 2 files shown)


zbeacon:
Hello @Sliosas,

Thank you for your interest in ThingsBoard IoT Gateway in Python.
We will review your commit and change the default converter for Modbus to read more than one register.

By the way, if you need any changes to processing the data in your project, please use custom converters. We are working with documentation for this feature. To connect a custom converter, you must place the file (the file name does not matter) together with your converter in the “extensions / modbus /” folder in the project folder (if you cloned the repository), and then add the “converter” ”” Specify the parameter in the section devices in the modbus .json file and give this parameter the class name of your custom converter.

For example:

...
"unitId": 1,
"DeviceName": "Temp Sensor",
"attributePollPeriod": 5000,
"timeseriesPollPeriod": 5000,
"converter": "MyCustomConverter",
"sendDataOnlyOnChange": true,
...

The user-defined init function converter will receive a device configuration section. And when the data is received, the function “convert” the converter will be called. A custom converter should return json data with a structure as follows:

{
"device_name": "Name1",
"deviceType": "ModbusDevice",
Telemetry: [{"telemetry_key1": "telemetry_data1"}],
"attribute": [{"attribute_key1": "attribute_data1"}]
}

In JSON, there can only be attributes or only telemetry. "deviceName" and "deviceType" are required.

PR #228, opened 2019-11-09
Bug fixes for Device Type and Extension issue to support if no filter expression is provided

ssmaurya:


zbeacon:
Hi, @ssmaurya , 
Improved functionality.

The new gateway doesn't uses filter expressions. You can try to use it instead of a Java gateway. 

PR #276, opened 2020-04-02
fix - can not load extension from /var/lib/thingsboard_gateway/extensions/

agtokty:
fix - can not load custom converters & connectors from /var/lib/thingsboard_gateway/extensions/


agtokty:
as stated [docs](https://thingsboard.io/docs/iot-gateway/custom/) , connectors that are placed in **/var/lib/thingsboard_gateway/extensions/serial** should load but, check_and_import functions looking in wrong folder :  **/var/lib/thingsboard_gateway/serial**

PR #282, opened 2020-04-07
Slibutti mqtt

slibutti:
Hi,
I performed a full cleanup of the MQTT connector; moreover, I added a couple handy options.

1 - It is now possible to specify mqtt QoS level for telemetry, attribute updates and so on
2 - It is now possible to ask for shared attributes values, instead of only subscribing for changes
3 - Code style is now more or less uniform in the whole file
4 - I inserted comments in the whole file
5 - Corrected a bug in disconnect method

You can test this version using a config file like this:

`
{
  "broker": {...},
  "mapping": [{..., "subscriptionQos": 1}],
  "connectRequests": [{..., "subscriptionQos": 1}],
  "disconnectRequests": [{..., "subscriptionQos": 1}],
  "attributeUpdates": [],
  "attributeRequests": [{
      "topicFilter": "up/attr/+/requests/+",
      "subscriptionQos": 1,
      "deviceNameTopicExpression": "(?<=attr\/)(.*?)(?=\/requests)",
      "attributeNameTopicExpression": "(?<=requests\/)(.*?)(?=$)",
      "topicExpression": "down/attr/${deviceName}/requests/${attributeKey}",
      "valueExpression": "{\"${attributeKey}\":\"${attributeValue}\"}"
    }],
  "serverSideRpc": []
}
`

Bests,
S

### DIFF
--- a/thingsboard_gateway/connectors/mqtt/mqtt_connector.py
+++ b/thingsboard_gateway/connectors/mqtt/mqtt_connector.py
@@ -36,13 +36,14 @@ class MqttConnector(Connector, Thread):
         self.__broker = config.get('broker')
         self.__mapping = config.get('mapping')
         self.__server_side_rpc = config.get('serverSideRpc', [])
-        self.__service_config = {"connectRequests": [], "disconnectRequests": []}
+        self.__service_config = {"connectRequests": [], "disconnectRequests": [], "attributeRequests" : []}
         self.__attribute_updates = config.get("attributeUpdates")
         self.__get_service_config(config)
 
         self.__mapping_sub_topics = {}
         self.__connect_sub_topics = {}
         self.__disconnect_sub_topics = {}
+        self.__attributes_sub_topics = {}
 
         client_id = ''.join(random.choice(string.ascii_lowercase) for _ in range(23))
         self._client = Client(client_id)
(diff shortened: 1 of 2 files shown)


thingsboard_gateway/connectors/mqtt/mqtt_connector.py
```diff
@@ -27,29 +27,72 @@
...
-        self.__mapping = config.get('mapping')
-        self.__server_side_rpc = config.get('serverSideRpc', [])
-        self.__service_config = {"connectRequests": [], "disconnectRequests": []}
-        self.__attribute_updates = config.get("attributeUpdates")
-        self.__get_service_config(config)
-        self.__sub_topics = {}
+
+        self.__mapping = [
+            x for x in config.get('mapping')
+            if x.get("topicFilter") and x.get("converter")
```
zbeacon, line 44: Please add here and in the same places logging like:
"topicFilter in config {...} - not found!"
"converter configuration not found in config: {}"
etc.

Logging will help to find a problem with configuration.

zbeacon:
Hi @slibutti 

Thank you for your work, please add more logging where the gateway reads configuration. You add check, for required parameters, but unfortunately, it will not show error to user.

thingsboard_gateway/connectors/mqtt/mqtt_connector.py
slibutti: That makes absolutely sense. Please check the last commit and let me know  it you like that.

PR #391, opened 2020-09-12
Master jsm

joerg555:
Hi, i am using thingsboard gateway OPC UA connector and i have fixed 2 problems.
Sample object struture:
"deviceNodePattern": "Root.Objects.Station_11001$"
"deviceNamePattern": "Station_${DeviceID}",
                "attributes": [
                    { "key": "DeviceID", "path": "${DeviceID}" }
                ],
                "timeseries": [
                    { "key": "temperature", "path": "${IO.Temp.0}" }
                ]

``
Root.Objects.Station_11001 (object)
Root.Objects.Station_11001.DeviceID (value)
Root.Objects.Station_11001.IO (object)
Root.Objects.Station_11001.IO.Temp (object)
Root.Objects.Station_11001.IO.Temp.0  (value)
`
1. The scanning of devices with "deviceNodePattern" works, but when you are in a sub-object "Root.Objects.Station_11001.IO"  tb gateway combines Root.Objects.Station_11001.Station_11001.IO and never find my "Station_11001.IO.Temp.0".
In OPC UA you cant concat namespaces, because they are unique.
The ns=2;s=Station_11001.IO of the "directory" should be a part of the Value ns, but must not be.
So normaly, you have to scan all sub-folters (objects) of the device. In my case now i search only:
new_node_path = 'Root.Objects.'+ str(new_node.nodeid.Identifier)
may be, we have make a parameter to search all sub-objects recursiv.

2. I had an exeption in get_value(), because the value was empty, i fixed this wtith try and catch, may be its an problem with your OPC UA library
In OPC UA, it is possible, that you have a value, that ist not messured. you can make a subscription, and will get the first value later, when its available 

3. I have optimised the code a little bit.

### DIFF
--- a/thingsboard_gateway/connectors/opcua/opcua_connector.py
+++ b/thingsboard_gateway/connectors/opcua/opcua_connector.py
@@ -288,7 +288,11 @@ class OpcUaConnector(Thread, Connector):
                 self.__search_node(device_info["deviceNode"], information_path, result=information_nodes)
                 for information_node in information_nodes:
                     if information_node is not None:
-                        information_value = information_node.get_value()
+                        try:
+                            information_value = information_node.get_value()
+                        except:
+                            log.error("Err get_value: %s", str(information_node))
+                            continue
                         log.debug("Node for %s \"%s\" with path: %s - FOUND! Current values is: %s",
                                   information_type,
                                   information_key,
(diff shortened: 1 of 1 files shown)


joerg555:
Hi thanks for your merge,
in my debug session i see a unknuwn parameter "deviceTypePattern" ist verry nice.
is the dokumentaion on github ? 
                "deviceNodePattern": "Root.Objects.Station_\\d+$",
                "deviceNamePattern": "Device_${DeviceID}",
                "deviceTypePattern": "MyType",
Regards Jörg

PR #398, opened 2020-09-24
 giusguerrini, 2020-09-24:

giusguerrini:
Bug Fix: if an OPC-UA source generates more than one value change in the same polling cycle, the platform receives only the  last one. Moreover, gateway's memory grows indefinitely.
Solution: flush event set and send all data to platform,
so data_to_send doesn't grow indefinitely in case of more than one value change
per cycle.
NOTE: possible performance improvement: use a map to store only one event per
variable to reduce frequency of messages to platform.


zbeacon:
Hi @giusguerrini ,

Thank you for your improvements, this PR will increase performance.

PR #413, opened 2020-10-11
Update opcua_connector.py

wgwhl:
when identity type is username, the config no  "type" parameter 
This part of configuration will look like:
    "identity": {
      "username": "user",
      "password": "5Tr0nG?@$sW0rD"
    },


wgwhl:
when identity type is username, the config no "type" parameter, so if self.__server_conf["identity"]["type"] == "cert.PEM": have a exception

PR #415, opened 2020-10-11
Update opcua_connector.py

wgwhl:
fixed two bug
1. when identity is username, if self.__server_conf["identity"]["type"] == "cert.PEM" have a exception.
   because This part of configuration will look like:
    "identity": {
      "username": "user",
      "password": "5Tr0nG?@$sW0rD"
    },
2.according to  description of node.get_parent(),sometime get_parent() return None, so sometime have a exception.


zbeacon:
Hi @wgwhl 

Thank you for your participating in project life.

Previous 2 PR were added, this will be closed.

PR #436, opened 2020-12-04
Update bytes_modbus_downlink_converter.py

arezola:
3 bug fixing added:
1. byte_order logic in line 30
2. variable size calculation in line 55 (change 8with a 16)
3. Fix line 99 with the logic needed for saving more than one register (32 and 64 bit variables)


arezola:
The proposed changes allow to save 32 and 64 bit variables correctly in modbus registers. Previous version only saved the first register.

PR #461, opened 2021-01-16
OPCUA new syntax for "${$DisplayName}", ... optimise OPCUA scan

joerg555:
Hello,
in OPC UA Gateway i have now new syntax "${$DisplayName}", ${$BrowseName}", "${$NodeId.Identifier}" to get devicename from current node without searching for OPC UA variable.
"mapping": [

    {
        "deviceNodePattern": "Root.Objects.Station_1300.U\\d+$",
        "deviceNamePattern": "Bostation_${$DisplayName}",
        "deviceTypePattern": "Bo",
        "attributes": [
            { "key": "Name", "path": "${Name}" }
        ],
        "timeseries": [
            { "key": "Wateralarm", "path": "${HwLes}" }
        ]
    }
]
This matches Root.Objects.Station_1300.U01, Root.Objects.Station_1300.U02, ...
Because the OPC UA Scan is verry slow (may be 300 ms for one tag) i optimise the scan.
My new function is "get_node_path(self, node)" to get the full path os a OPC UA node. The old one node.get_path() is verry slow and dose not Work with deep pathes (e.g. "root.objects.heater1.pump1.value"). 
I have a question about configrations with "\\\\.", "\.", "\\.". what is the exact syntax for it ?
Could you please check my changes with OPC UA servers with views.
My test platform has 200 devices with 20000 variables.
keep clean Jörg

### DIFF
--- a/thingsboard_gateway/connectors/opcua/opcua_connector.py
+++ b/thingsboard_gateway/connectors/opcua/opcua_connector.py
@@ -394,17 +394,25 @@ class OpcUaConnector(Thread, Connector):
                 name_expression = TBUtility.get_value(name_pattern_config, get_tag=True)
                 if "${" in name_pattern_config and "}" in name_pattern_config:
                     log.debug("Looking for device name")
-                    name_path = self._check_path(name_expression, device_node)
-                    device_name_node = []
-                    self.__search_node(device_node, name_path, result=device_name_node)
-                    device_name_node = device_name_node[0]
-                    if device_name_node is not None:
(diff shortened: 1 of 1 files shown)


zbeacon:
Hi @joerg555 ,

Thank you for your participating in project life, according to your question: "
I have a question about configrations with "\\.", ".", "\.". what is the exact syntax for it ?"

The correct separating symbol is "\\.", because we use this string to looking for node on OPC-UA server using regular expressions, so we should screen dots.

PR #479, opened 2021-02-17
Update to fix and Update "SendDataOnlyOnChange" behaviour

arezola:
- This pull request will overwrite a previous pending pull request: "resolve bug for timeseries in modbus when sendDataOnlyOnChange is active #477"
- SendDataOnlyOnChange logic update:
    -- Code changes fix the "SendDataOnlyOnChange" bug for telemetry data.
    -- Code changes to allow including "SendDataOnlyOnChange" parameter within each variable. 
    -- This version allows to define a general value of "SendDataOnlyOnChange" and in addition a particular behaviour of "SendDataOnlyOnChange" for particular variables, i.e. we can define "SendDataOnlyOnChange" as True for all the variables and then if there is a variable where we need to set "SendDataOnlyOnChange" to False it can be included as a parameter within such variable (attribute or telemetry) in the configuration Json.


arezola:
This pull request is not correct. Sorry for the inconveniences.

PR #480, opened 2021-02-18
Request connector bugfix.

dravai:
Indentation problem in request connector fixed. There is no need to call TBUtility.get_value() for keys and should check "ts" tag in response json.

### DIFF
--- a/thingsboard_gateway/connectors/request/json_request_uplink_converter.py
+++ b/thingsboard_gateway/connectors/request/json_request_uplink_converter.py
@@ -40,10 +40,13 @@ class JsonRequestUplinkConverter(RequestConverter):
             log.exception(e)
 
         try:
+            has_timestamp = False
             for datatype in self.__datatypes:
                 current_datatype = self.__datatypes[datatype]
                 for datatype_object_config in self.__config["converter"].get(datatype, []):
-                    datatype_object_config_key = TBUtility.get_value(datatype_object_config["key"], data, datatype_object_config["type"], expression_instead_none=True)
+                    datatype_object_config_key = datatype_object_config["key"]
+                    if datatype_object_config_key == "ts":
+                        has_timestamp = True
                     datatype_object_config_value = TBUtility.get_value(datatype_object_config["value"], data, datatype_object_config["type"])
                     if datatype_object_config_key is not None and datatype_object_config_value is not None:
                         dict_result[current_datatype].append({datatype_object_config_key: datatype_object_config_value})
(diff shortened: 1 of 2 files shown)


zbeacon (changes requested):
Hi @dravai ,

Thank you for your participating in project life, but we have to process telemetry and attributes using TBUtility.get_value method to keep compatibility with JSONPath expressions in configuration for keys, instead of getting key from result.

dravai:
Hi @zbeacon ,

I have the following configuration
`        "telemetry": [
          {
            "key": "ts",
            "type": "long",
            "value": "${scan_at}"
          },
          {
            "key": "sap_order_number",
            "type": "string",
            "value": "${worksheet.sap_order_number}"
          },
          {
            "key": "item_number",
            "type": "string",
            "value": "${product.cikkszam}"
          },
          {
            "key": "quantity",
            "type": "long",
            "value": "${quantity}"
          }
        ]
`
and data from REST request
`{
  'scan_at': 1613980933093, 
  'quantity': 1, 
  'worksheet': {
    'sheet_number': 'SN-09859', 
    'sap_order_number': 'SAP-ON/9746139133', 
   }, 
  'product': {
    'cikkszam': 'CSZ-6322285798', 
    'megnevezes': 'Accessory mattresses time minuses process', 
  }
}
`
TB_Utility.get_value for **quantity** key returns the value of it, instead of the name. In this case return value should be 'quantity' but you get 1. Could you tell me what should I modify in the config to return correct name of quantity key?

Thanks for your help.

zbeacon:
> Hi @zbeacon ,
> 
> I have the following configuration
> ` "telemetry": [ { "key": "ts", "type": "long", "value": "${scan_at}" }, { "key": "sap_order_number", "type": "string", "value": "${worksheet.sap_order_number}" }, { "key": "item_number", "type": "string", "value": "${product.cikkszam}" }, { "key": "quantity", "type": "long", "value": "${quantity}" } ]`
> and data from REST request
> `{ 'scan_at': 1613980933093, 'quantity': 1, 'worksheet': { 'sheet_number': 'SN-09859', 'sap_order_number': 'SAP-ON/9746139133', }, 'product': { 'cikkszam': 'CSZ-6322285798', 'megnevezes': 'Accessory mattresses time minuses process', } } `
> TBUtility.get_value for **quantity** key returns the value of it, instead of the name. In this case return value should be 'quantity' but I've got 1. Could you tell me what should I modify in the config to get correct name of quantity key?
> 
> Thanks for your help.

Hi @dravai ,

I agree with you, in this case (When you want to use the same key like in the message) TBUtility.get_value using is useless. But it also breaks a case when you want to use value of some received data as a key. Probably we should use converting like in the MQTT uplink converter. What do you think about this?

PR #520, opened 2021-05-12
Bugfix: validation always fails: 'str' object has no attribute 'get'

slibutti:
Recently changed data validation function does its checks by treating
attributes/values as they were dictionaries; however, they are lists
of dictionaries.

data["telemetry"].get("ts") is bound to fail: the correct check would be
data["telemetry"][i].get("ts"), for each element i.

Signed-off-by: Simone Libutti <simone.libutti@ibtsystems.it>

### DIFF
--- a/thingsboard_gateway/tb_utility/tb_utility.py
+++ b/thingsboard_gateway/tb_utility/tb_utility.py
@@ -43,8 +43,26 @@ class TBUtility:
             error = 'deviceName is empty in data: '
         if error is None and not data.get("deviceType"):
             error = 'deviceType is empty in data: '
-        if error is None and data.get("attributes") is None and (data.get("telemetry") is None or (data["telemetry"].get("ts") is not None and len(data["telemetry"].get("values")) == 0)):
-            error = 'No telemetry and attributes in data: '
+
+        if error is None:
+            got_attributes = False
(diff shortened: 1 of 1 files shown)


thingsboard_gateway/tb_utility/tb_utility.py
```diff
@@ -43,8 +43,26 @@ def validate_converted_data(data):
...
-        if error is None and data.get("attributes") is None and (data.get("telemetry") is None or (data["telemetry"].get("ts") is not None and len(data["telemetry"].get("values")) == 0)):
-            error = 'No telemetry and attributes in data: '
+
+        if error is None:
+            got_attributes = False
+            got_telemetry = False
+
+            if data.get("attributes") is not None:
+                for entry in data.get("attributes"):
+                    if entry.get("ts") is not None and len(entry.get("values")) > 0:
```
zbeacon, line 53: Attributes doesn't have an ability to contain timestamp in the message (According to ThingsBoard architecture). This check will be redundant because we always have attributes just like a dictionary. Could you change this?

slibutti: Fixed it. I'm partial to leave the len check on attributes as it is; let me know if you want it removed.

PR #538, opened 2021-07-22
Source distribution fixes

pbrkr:
These changes ensure that the example config files and the license text are included in the source distribution uploaded to pypi. They were tested by running `python3 setup.py sdist` and examining the generated tarball.

I recommend making a minor release if these changes are accepted so that the new files show up in a release on pypi.


pbrkr:
@zbeacon ping?

PR #548, opened 2021-08-07
Gateway RPC fixes

mjbernet3:
1) Fixes connector type issue in multiple connectors
2) Removes module from RPC method earlier so the module is not sent to a gateway's server side rpc handler

Explanation of first fix:

Within _rpc_request_handler in the TBGatewayService, a connector's type is compared to the module initially attached to a method to determine whether to forward the content to the server side rpc handler.  Without the correct property "_connector_type" specified on a connector, you will receive an error saying "[Connector Class] object has no attribute _connector_type."  Below is the relevant code from the _rpc_request_handler.

tb_gateway_service

Explanation of second fix:

Before this change the content was passed to the server side rpc handler from _rpc_request_handler without removing the module from the method.  This is not that big of a deal, but I figured it would be cleaner to remove the module from the method before passing the content so there is no need to remove it from a connector.

Let me know if there are any issues with this :)


thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -529,7 +530,7 @@ def __rpc_gateway_processing(self, request_id, content):
         arguments = content.get('params', {})
         if content.get("timeout") is not None:
             arguments.update({"timeout": content["timeout"]})
-        method_to_call = content["method"].replace("gateway_", "")
+        method_to_call = content["method"]
```
zbeacon, line 533: This part of the code relates to RPC request to the gateway device and we pass the full message to this function, so we should remove "gateway_" prefix. Please take a look what contains "content" variable.

zbeacon:
Hi @mjbernet3 ,

Thank you for your participating. I agree with the first issue, but the second one is not an issue. It is by design and like in the documentation of the gateway's features. We set it to make some visual separation on ThingsBoard dashboard. You are right it is not necessary in the code and you can use methods without this prefix, but it allows to simplify understanding what methods and dashboards relate to the gateway and what are relate to the devices for beginners.

thingsboard_gateway/gateway/tb_gateway_service.py
mjbernet3: So are you saying that the problem with this change is that the "gateway_" prefix is no longer removed?  What I wanted to achieve with this change was to remove the module from the content's method earlier so that it would not be passed to the server_side_rpc_handler of a gateway.  Unless I am mistaken, the "gateway_" prefix is still removed.  

For example, if the method of the content started as "gateway_ping", then the method would be split on line 498 into "gateway" and "ping."  Then on line 502, I set the method of the content to "ping."  Therefore, when content is passed to __rpc_gateway_processing, the prefix is already removed.

If this change is still not wanted let me know and I will remove it so that only the connector type issue is fixed.  Sorry for the delayed response, school has been very busy.

PR #572, opened 2021-09-20
Feature: SQlite Storage 

TheSamabo:
Hi, 

First of all this is my first time contributing to any open source project. Its was a lot of fun making something that we would use in production. So i wanted to give back.

What my contribution features:
- New sqlite3 storage type for outgoing messages.
- New gateway RPC call to retrieve data for a specific connected device.
- Thread safe inserts and reads from sqlite storage.
- New separate log file for database
- Tidiness in all storage type's files/folders

## Overview

File based storage is very convenient, but it has some major flaws that i could not tolerate. Like unreliable data restore after internet blackout. Here i will explain how this storage works and give you some high level overview.

There are 3 major components:
1. Database Connector
2. Database
3. Storage Handler       

***Database connector*** provides low level interactions with database (commits, connection, rollbacks, db closing)
In the main tb_gateway.yaml config, you need to specify **data_file_path** so that the connector can create the database file

***Database*** Executes all the necessary sql queries to store data reliably. 
Data diagram is quite simple:

connected_devices(
		deviceName TEXT,  <- Stores device Name
		deviceType TEXT,	<- stores device type  *not requeired
		connector TEXT,		 <- connector name 
		dataSavedIndex INTEGER, <- These fields are to store index of the message		             dataUploadedIndex INTEGER  that was stored and index of message sent to TB
		)  

**Device Table:**
_AAAAAAAAAA <- device name first 10 chars hashed SHA1 and capitalized
(
dataIndex PRIMARY KEY INTEGER,  
timestamp INTEGER   <-  timestamp from the message
message TEXT        	<- JSON encoded message
)

When a connector or gateway service calls ***add_device()*** a new entry is added to table "connected_devices" also
a new table is created from the device's name. Device's name gets run through SHA1 hashing algorithm then capitalized and finally prefixed with symbol:  ***_***  . 

This component's main method **process()** takes item in a **processQueue** and unwraps it from **DatabaseRequest** class. Where it finds what operation and sql statement should be executed depending on the **DatabaseActionType** 

***Storage Handler***  Is high level component to this storage type. It  encapsulates needed method like **.put()** 

## RPC Call
I have implemented a new gateway specific RPC call to retrieve data from sqlite storage.
There are some parameters that are necessary for this to work. 

1. Device Name
2. Timestamp from which to start reading

So the params will look like this: "{"deviceName": "\<deviceName\>", "ts": \<unixTimestamp\>}"

Method to call: **gateway_data**

Right now you can only specify a timestamp from which we start reading until present moment is reached.

## Data retention
I know data cannot be stored on device forever, so i implemented a data retention mechanism. That deletes data from all device tables that is older than a specified number of days in tb_gateway.yaml config

for example:
if i set **max_days_to_store_data** to *7*, then any messages older then 7 days will be deleted

This clean up mechanism is right now hard coded to execute every 10 messages processed. This could be later changed to a value configurable in tb_gateway.yaml

***

Please let me know what do think of these changes. I know there are things that need to be improved, but i wanted to show my work so that we can have a discussion about some of the changes and improvments.

Thank you,
Sam


TheSamabo:
Uhh, i bit of a problem.
I can't sign the CLA with my gitlab account.
I am solo author of this PR. 
Both the TheSamabo and SamuelRuzek are my accounts.

PR #613, opened 2021-11-01
Fixed higt load CPU usage

samson0v:


thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -417,6 +417,7 @@ def __send_to_storage(self):
                         log.error('Data from the device "%s" cannot be saved, connector name is %s.',
                                   data["deviceName"],
                                   connector_name)
+                sleep(0.2)
```
zbeacon, line 420: Should we sleep if have a lot of data here?

PR #621, opened 2021-11-04
Added service RPC methods for OPCUA Connector

samson0v:


thingsboard_gateway/connectors/opcua/opcua_connector.py
```diff
@@ -222,10 +222,58 @@ def on_attributes_update(self, content):
...
+                except IndexError:
+                    log.error('Not enough arguments. Expected min 2.')
+                    self.__gateway.send_rpc_reply(content['device'],
+                                                  content['data']['id'],
+                                                  {content['data']['method']: 'Not enough arguments. Expected min 2.',
+                                                   'code': 400})
+
+                node_list = []
+                self.__search_node(current_node=content['device'], fullpath=f'ns={ns};i={i}', result=node_list)
+
```
zbeacon, line 240: Node identifier can be not only i, it may be also like a path or with different type:

```
i | NUMERIC (UInteger)
s | STRING (String)
g | GUID (Guid)
b | OPAQUE (ByteString)
```

PR #626, opened 2021-11-11
Added argumets for SQLite Storage

samson0v:


thingsboard_gateway/storage/sqlite/database.py
```diff
@@ -95,6 +104,8 @@ def process(self):
 
                         self.db.commit()
 
+                    sleep(.2)
```
zbeacon, line 107: Probably we have to sleep when we don't process any data instead of sleeping after every processing, what do you think?

PR #653, opened 2021-12-16
BLE Connector Remastered

samson0v:


thingsboard_gateway/connectors/ble/device.py
```diff
@@ -0,0 +1,249 @@
...
+
+    async def timer(self):
+        await self.__process_self()
+        self.last_polled_time = time()
+
+        while True:
+            if time() - self.last_polled_time >= self.poll_period:
+                self.last_polled_time = time()
+                await self.__process_self()
+
```
zbeacon, line 93: Probably it will be better to put sleep into else statement

thingsboard_gateway/connectors/ble/device.py
```diff
@@ -0,0 +1,249 @@
...
+                        'config': {
+                            'attributes': self.config['attributes'],
+                            'telemetry': self.config['telemetry']
+                        },
+                        'data': not_converted_data
+                    }
+
+                    self.callback(data_for_converter)
+
+                    return
```
zbeacon, line 115: What if we want to read one characteristic as a telemetry and an attribute keys?

PR #704, opened 2022-01-19
Added reconnecting functionality for BLE Device

samson0v:


thingsboard_gateway/connectors/ble/device.py
```diff
@@ -151,7 +164,7 @@ async def __process_self(self):
                         self.__set_char_handle(item, char_id)
                         self.notifying_chars.append(char_id)
                         await self.notify(char_id)
-                    except BleakError:
+                    except BleakError as e:
                         log.error(e)
```
zbeacon, line 168: What if we receive an error during device reading? Could we schedule a reconnect and reading/ subscribing for notification after receiving an error?

thingsboard_gateway/connectors/ble/device.py
```diff
@@ -105,11 +105,24 @@ async def timer(self):
...
+
+                try:
+                    await self.client.disconnect()
+                except Exception as err:
+                    log.exception(err)
+
+                while not self.stopped and not self.client.is_connected:
+                    await self.connect_to_device()
+
+                    sleep(.2)
```
zbeacon, line 125: What if we cannot connect to device again? How  many tries the gateway will do and how often?

thingsboard_gateway/config/ble.json
```diff
@@ -13,6 +13,7 @@
       "pollPeriod": 500000,
       "showMap": false,
       "timeout": 10000,
+      "connectRetry": 5,
       "telemetry": [
```
zbeacon, line 17: Please add something like "wipeConnectionTriesInSeconds" and set connect try to 0 after this time. It will give the ability to reconnect to device after some period.

PR #717, opened 2022-01-31
Added opportunity to receive different data types for REST Connector

samson0v:


thingsboard_gateway/connectors/rest/rest_connector.py
```diff
@@ -302,57 +302,73 @@ def __send_request(self, request_dict, converter_queue, logger, with_queue=True)
...
+    def endpoint(self):
+        return self.__endpoint
+
+    @staticmethod
+    async def _convert_data_from_request(request):
+        try:
+            json_data = await request.json()
+        except json.decoder.JSONDecodeError:
+            data = await request.post()
+            json_data = json.dumps(dict(data))
```
zbeacon, line 325: What if we receive not an object in request? E.g. bytes/string/etc.

PR #723, opened 2022-02-02
Added functionality for checking devices idle time

samson0v:


everydaycodings:
#### Have You Tried This One?
https://stackoverflow.com/questions/911856/detecting-idle-time-using-python

Issue #735, opened 2022-02-08
[BUG] OPC UA connector - telemetry and attribute mapping fails


ysimonx:
**Matching keys with regex fails **
in my opcua.json, i have a key for both telemetry and attributes
${User Registers\\.cPURE\\.Fan\\.DeltaP_100}

but, opc ua connector matches this key with both

User Registers\.cPURE\.Fan\.DeltaP_10
and
User Registers\.cPURE\.Fan\.DeltaP_100

because of that

line 589 opcua connector ...  

`                        regex_search = fullpath_pattern.fullmatch(nnp1, partial=True) or \
                                       nnp2 in full1 or \
                                       nnp1 in full1
`

partial=True should not be applied for a variable

** OPC-UA Connector:**
when removing the line # 601
`
result.append(child_node)
`

telemetry and attributes are ok

**Versions 3.0.1 - last python source edition from github :**
 - OS: mac OS
 - Thingsboard IoT Gateway version 3.0.1
 - Python version 3.9

ysimonx:
cf : https://github.com/thingsboard/thingsboard-gateway/pull/736

PR #736, opened 2022-02-08
opc ua : mapping telemetry and/or attributes fails

ysimonx:
cf https://github.com/thingsboard/thingsboard-gateway/issues/735

in my opcua.json, i have a key for both telemetry and attributes
${User Registers\.cPURE\.Fan\.DeltaP_100}

but, opc ua connector matches this key with both

User Registers.cPURE.Fan.DeltaP_10
and
User Registers.cPURE.Fan.DeltaP_100

we should not select this node (DeltaP_10 !!), 
this node is selected because the "scanning" process selects node of type variable, even when a partial match is done.

### DIFF
--- a/thingsboard_gateway/connectors/opcua/opcua_connector.py
+++ b/thingsboard_gateway/connectors/opcua/opcua_connector.py
@@ -596,9 +596,9 @@ class OpcUaConnector(Thread, Connector):
                                 if self.__show_map:
                                     log.debug("SHOW MAP: Search in %s", new_node_path)
                                 self.__search_node(child_node, fullpath, result=result)
-                            elif new_node_class == ua.NodeClass.Variable:
-                                log.debug("Found in %s", new_node_path)
-                                result.append(child_node)
+                            # elif new_node_class == ua.NodeClass.Variable:
+                            #     log.debug("Found in %s", new_node_path)
+                            #     result.append(child_node)
                             elif new_node_class == ua.NodeClass.Method and search_method:
                                 log.debug("Found in %s", new_node_path)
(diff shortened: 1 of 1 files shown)


ysimonx:
@zbeacon : may be you can merge this one too ;-)

how long time will it takes before docker image will be updated with the master branch please ?

zbeacon:
@ysimonx ,

I want to test it with regex, this change may cause the issue with regex multi search for nodes.

Docker usually updates on release. But there are Docker file in docker folder, you can copy it to the root and run the following command to build your own docker image for current architecture:
`sudo docker build .`
At the end you will receive an id of the image and will be able to run it with the following command:
`sudo docker run YOUR_IMAGE_ID`

ysimonx:
[zbeacon](https://github.com/zbeacon)

here are some screenshots

from opcua.json

Capture d’écran 2022-02-10 à 22 33 05

Capture d’écran 2022-02-10 à 22 33 32

and here, what the opcua server sees as subscription

Capture d’écran 2022-02-10 à 22 34 06

as you can see : delta_p10 and timer are subscribed, although i did not configure them in opcua.json

lines 591 to 594 are used for every variable i configured
```
                    regex_fullmatch = regex.fullmatch(fullpath_pattern, nnp1) or \
                                      nnp2 == full1 or \
                                      nnp2 == fullpath or \
                                      nnp1 == full1
```

but 
lines 600 to 602 should be used only for recursive browsing
```
                        regex_search = fullpath_pattern.fullmatch(nnp1, partial=True) or \
                                       nnp2 in full1 or \
                                       nnp1 in full1
```

PR #761, opened 2022-03-01
Changed code to handle null value for releasing priority

Kretiss:


Kretiss:
@zbeacon Hi,

When do you plan to release new version?

PR #788, opened 2022-03-25
Added Grpc Socket Connector

samson0v:


thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -44,6 +47,7 @@
 try:
     from thingsboard_gateway.gateway.grpc_service.grpc_connector import GrpcConnector
     from thingsboard_gateway.gateway.grpc_service.tb_grpc_manager import TBGRPCServerManager
+
     GRPC_LOADED = True
```
imbeacon, line 51: Please make sure that the gateway will work even without installed GRPC (On some platforms like OpenWRT - GRPC is not supported).

PR #815, opened 2022-04-26
fix SimpleQueue not found with python < 3.7

ninhnn2:


samson0v:
Hi @ninhnn2, you can see from setup.py that we don't support the version of python < 3.7.
Also, official support was ended for python 3.6.

imbeacon:
Hi @ninhnn2 ,

Thank you for participating in the project life, but as told @samson0v, the basic version of the python, required by the gateway is at least 3.7. So these changes for queues are not required. By the way, queues used not only in the general service, they also used in some connectors like request connector, and these changes may produce some additional bugs and errors.

PR #817, opened 2022-05-04
Fix ignoring more than one command configurations.

devaskim:
Without this PR multiple command configuration (see example below) for the same nodeId doesn't work at all:
```
            "timeseries":
            [ {
                    "key": "INVCH_Bdc_bExporting",
                    "nodeId": 419419328,
                    "isExtendedId": true,
                    "command": "0:1:little:0",
                    "value": "2:1:int",
                    "expression": "value"
                }, {
                    "key": "18FFD4C0_Index3_Byte1",
                    "nodeId": 419419328,
                    "isExtendedId": true,
                    "command": "0:1:little:3",
                    "value": "1:1:int",
                    "expression": "value"
                }
            ]
        }
```

### DIFF
--- a/thingsboard_gateway/connectors/can/can_connector.py
+++ b/thingsboard_gateway/connectors/can/can_connector.py
@@ -439,7 +439,7 @@ class CanConnector(Connector, Thread):
                                     self.get_name(), tb_key, config_key, )
                         continue
 
-                    if msg_config.get("command", ""):
+                    if msg_config.get("command"):
                         cmd_config = self.__parse_command_config(msg_config["command"])
                         if cmd_config is None:
                             log.warning("[%s] Ignore '%s' %s configuration: wrong command configuration",
@@ -451,8 +451,14 @@ class CanConnector(Connector, Thread):
                         if node_id not in self.__commands:
                             self.__commands[node_id] = cmd_config
                         else:
-                            # TODO: warn user if previous cmd_config differs from current cmd_config
-                            pass
+                            prev_cmd_config = self.__commands[node_id]
+                            if cmd_config["start"] != prev_cmd_config["start"] or \
+                                    cmd_config["length"] != prev_cmd_config["length"] or \
+                                    cmd_config["byteorder"] != prev_cmd_config["byteorder"]:
+                                log.warning("[%s] Ignore '%s' %s configuration: "
+                                            "another command configuration already added for arbitration_id %d",
+                                            self.get_name(), tb_key, config_key, node_id)
+                                continue
                     else:
                         cmd_id = self.NO_CMD_ID
                         self.__commands[node_id] = None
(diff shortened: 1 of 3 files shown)


apadEnlace:
Hi!

Any updates on when this fix could be implemented and released to the users?

Thanks for the support!

devaskim:
@apadEnlace Could you confirm that this PR fixes your case? Maybe this speed up any feedback from Gateway team - @imbeacon and @samson0v

thingsboard_gateway/connectors/can/can_connector.py
```diff
@@ -439,15 +439,20 @@ def __parse_config(self, config):
                                     self.get_name(), tb_key, config_key, )
                         continue
 
-                    if msg_config.get("command", "") and node_id not in self.__commands:
+                    if msg_config.get("command", ""):
```
samson0v, line 442: CHANGE TO if msg_config.get("command"):

thingsboard_gateway/connectors/can/can_connector.py
```diff
@@ -439,15 +439,20 @@ def __parse_config(self, config):
...
                             continue
 
                         cmd_id = cmd_config["value"]
-                        self.__commands[node_id] = cmd_config
+
+                        if node_id not in self.__commands:
+                            self.__commands[node_id] = cmd_config
+                        else:
+                            # TODO: warn user if previous cmd_config differs from current cmd_config
+                            pass
```
samson0v, line 455: Complete this code block

samson0v:
Hi @devaskim, thanks for your PR and interest in ThingsBoard IoT Gateway!
Resolve the comments above and we will merge it.

thingsboard_gateway/connectors/can/can_connector.py
devaskim: Done

thingsboard_gateway/connectors/can/can_connector.py
devaskim: Added check and log message with warning level. Also added unit test for both scenarious.

devaskim:
@samson0v All jobs done, even more - I added unit tests)))

By the way, do you know is [Thingsboard Dart client](https://github.com/thingsboard/dart_thingsboard_client) project alive? There is important (I think so) fixes - mine))) and other developer, and no feedback for a long time

samson0v:
@devaskim I will know if the dart team is alive and let you know)

samson0v:
@devaskim can you provide more info about your problem with the mobile client?

devaskim:
> @devaskim can you provide more info about your problem with the mobile client?

The problem is what some of entity relation API is broken in the client. Here are [PRs](https://github.com/thingsboard/dart_thingsboard_client/pulls) with more details.

samson0v:
@devaskim thank you very much, I will deliver your problem to the team

devaskim:
@samson0v What should I do now in scope of this PR?

PR #833, opened 2022-05-24
Small bugfix. Add opcua_asyncio/__init__.py extension file to avoid e…

JavierNR:
…rror with setup.py


pbrkr:
@imbeacon: Can this be backported to the release/3.1 branch? The file is also missing there and so neither v3.1 or the HEAD of release/3.1 builds.

PR #837, opened 2022-05-27
thingsboard-gateway was failing to call __send_data_pack_to_storage

KevinPoole:
The logic for sending large payloads in pieces was flawed.

1. If there was a single key/value and it was over the size limit by itself it would never get sent since the size check was at the start of the loop before it had ever been added.
2. If there were any unsent-adopted things after the loops (as in, a send had occurred and then more things were adopted without hitting the size limit) then these were never sent either.

Fixed these two issues and added some light commenting to clarify what is going on.


KevinPoole:
Going to re-attempt with a properly linked account...

PR #838, opened 2022-05-27
thingsboard-gateway was failing to call __send_data_pack_to_storage

KevinPoole:
The logic for sending large payloads in pieces was flawed.

1. If there was a single key/value and it was over the size limit by itself it would never get sent since the size check was at the start of the loop before it had ever been added.

2. If there were any unsent-adopted things after the loops (as in, a send had occurred and then more things were adopted without hitting the size limit) then these were never sent either.

Fixed these two issues and added some light commenting to clarify what is going on.

### DIFF
--- a/thingsboard_gateway/gateway/tb_gateway_service.py
+++ b/thingsboard_gateway/gateway/tb_gateway_service.py
@@ -634,23 +634,25 @@ class TBGatewayService:
 
                         max_data_size = self.__config["thingsboard"].get("maxPayloadSizeBytes", 1024)
                         if self.__get_data_size(data) >= max_data_size:
+                            # Data is too large, so we will attempt to send in pieces
                             adopted_data = {"deviceName": data['deviceName'],
                                             "deviceType": data['deviceType'],
                                             "attributes": {},
                                             "telemetry": []}
+
+                            # First, loop through the attributes
                             for attribute in data['attributes']:
+                                adopted_data['attributes'].update({attribute: data['attributes'][attribute]})
                                 adopted_data_size = self.__get_data_size(adopted_data)
                                 if adopted_data_size >= max_data_size:
+                                    # We have surpassed the max_data_size, so send what we have and clear attributes
                                     self.__send_data_pack_to_storage(adopted_data, connector_name)
                                     adopted_data['attributes'] = {}
-                                adopted_data['attributes'].update({attribute: data['attributes'][attribute]})
+                            
+                            # Now, loop through telemetry. Possibly have some unsent attributes that have been adopted.
                             for ts_kv_list in data['telemetry']:
                                 ts = ts_kv_list['ts']
                                 for kv in ts_kv_list['values']:
-                                    adopted_data_size = self.__get_data_size(adopted_data)
-                                    if adopted_data_size >= max_data_size:
-                                        self.__send_data_pack_to_storage(adopted_data, connector_name)
-                                        adopted_data['telemetry'] = []
                                     if len(adopted_data['telemetry']) == 0:
                                         adopted_data['telemetry'] = [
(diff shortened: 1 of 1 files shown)


thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -634,23 +634,25 @@ def __send_to_storage(self):
...
                         if self.__get_data_size(data) >= max_data_size:
+                            # Data is too large, so we will attempt to send in pieces
                             adopted_data = {"deviceName": data['deviceName'],
                                             "deviceType": data['deviceType'],
                                             "attributes": {},
                                             "telemetry": []}
+
+                            # First, loop through the attributes
                             for attribute in data['attributes']:
+                                adopted_data['attributes'].update({attribute: data['attributes'][attribute]})
```
samson0v, line 645: `adopted_data['attributes'].update(attribute)`

`attribute` - dict object
`data['attributes']`is a list and you want to get the item from this list by dict!
Also, we want in this place simply update attributes in adopted data, so can update by attribute without extra manipulations.

thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -634,23 +634,25 @@ def __send_to_storage(self):
...
                                 adopted_data_size = self.__get_data_size(adopted_data)
                                 if adopted_data_size >= max_data_size:
+                                    # We have surpassed the max_data_size, so send what we have and clear attributes
                                     self.__send_data_pack_to_storage(adopted_data, connector_name)
                                     adopted_data['attributes'] = {}
-                                adopted_data['attributes'].update({attribute: data['attributes'][attribute]})
+                            
+                            # Now, loop through telemetry. Possibly have some unsent attributes that have been adopted.
                             for ts_kv_list in data['telemetry']:
```
samson0v, line 653: Before this line it has to be the next line:
`telemetry = data['telemetry'] if isinstance(data['telemetry'], list) else [data['telemetry']]`

Because see on method `__convert_telemetry_to_ts` - it can return dict and list (so if it is dict we have the exception in the below lines where we want to get ts from current iter dict.

samson0v, line 653: Before this line it has to be the next line:
`telemetry = data['telemetry'] if isinstance(data['telemetry'], list) else [data['telemetry']]`

Because see on method `__convert_telemetry_to_ts` - it can return dict and list (so if it is dict we have the exception in the below lines where we want to get ts from current iter dict.

thingsboard_gateway/gateway/tb_gateway_service.py
KevinPoole: @samson0v I will be honest, I am having a slightly hard time understanding your feedback in the review here.

All I did was to move this line you have commented on from after the size check to before it.

Are you suggesting that the line which is currently:
```
adopted_data['attributes'].update({attribute: data['attributes'][attribute]})
```
needs to be changed to:
```
adopted_data['attributes'].update(attribute)
```
?

samson0v: Yes

thingsboard_gateway/gateway/tb_gateway_service.py
KevinPoole: and here are you saying that the line:
```
telemetry = data['telemetry'] if isinstance(data['telemetry'], list) else [data['telemetry']]
```
must be added before the loop and that the loop would become:
```
for ts_kv_list in telemetry:
```
?

KevinPoole: I had not seen your second comment when I replied. I understand now, thank you!

samson0v: Yes

PR #869, opened 2022-07-05
Serial connector: Load device config from device array 

anirudh-mehta:
**Description:** 

The PR resolves three issues: 
- [x] The current implementation for custom serial connector looks for the device configuration at the root of the configuration json. As per the [documentation](https://thingsboard.io/docs/iot-gateway/custom/#step-1-define-serialconnector-configuration) and to support configurability at the device level, the configuration should be fetched from each device array item. 
- [x] The thingsboard gateway's **_del_device_** method expects the device name as an input. The current implementation wrongly passes the device object instead.  
- [x] Install pyserial library is not installed

### DIFF
--- a/thingsboard_gateway/extensions/serial/custom_serial_connector.py
+++ b/thingsboard_gateway/extensions/serial/custom_serial_connector.py
@@ -18,8 +18,14 @@ import time
 from random import choice
 from string import ascii_lowercase
 from threading import Thread
+from thingsboard_gateway.tb_utility.tb_utility import TBUtility
 
-import serial
+try:
+    import serial
+except ImportError:
+    print("pyserial library not found - installing...")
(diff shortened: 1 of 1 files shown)


imbeacon:
Hi @anirudh-mehta ,

thank you for your update, but in this case configuration from the official site won't work, please add a case with only one configuration, you can check the type of config and if it contains only one object, put it into the array and pass to your logic.

anirudh-mehta:
@imbeacon can you please point me to the official site version you are referring to - I will update accordingly? However, I referred the one in ["IoT Gateway Customization Guide" ](https://thingsboard.io/docs/iot-gateway/custom/) and that uses the device array configuration. 

image

imbeacon:
I'm sorry, my bad. Thanks for participating, your changes merged.

PR #940, opened 2022-10-07
Explicitly set maxPayloadSizeBytes in gateway configuration.

devaskim:
It would be great to re-check if the others configuration parameters are present in *tb_gateway.yaml*


thingsboard_gateway/config/tb_gateway.yaml
```diff
@@ -7,6 +7,7 @@ thingsboard:
     enable: true
     statsSendPeriodInSeconds: 3600
     configuration: statistics.json
+  maxPayloadSizeBytes: 1000
```
samson0v, line 10: set to 1024

devaskim: Done

PR #944, opened 2022-10-11
Move device name and type extracting method to utility class.

devaskim:
These methods are useful in custom converters as well so I suggest to shared them through utility class.


samson0v:
@devaskim, We think that it will be better to leave these methods in MQTT Connector as they are `staticmethods` so in a custom converter you can simply import MQTTConverter and use them.

thingsboard_gateway/connectors/mqtt/mqtt_converter_utils.py
```diff
@@ -0,0 +1,69 @@
...
+
+from thingsboard_gateway.tb_utility.tb_utility import TBUtility
+
+log = getLogger("service")
+
+
+class MqttConverterUtils:
+
+    @staticmethod
+    def create_result_item(key, value, timestamp):
```
samson0v, line 27: Move these methods to `json_mqtt_uplink_converter` as `staticmethods`.

devaskim:
@samson0v Maybe I am not thinking as Python-way developer but from the point of view based on class hierarchy it is a bit strange to reuse method implemented in a derived class (`json_mqtt_uplink_converter `) in other derived class (my custom converter). I think such methods should be either part of parent class or part of utilities collection. Are you agree with me or maybe I didn't understand your explanation?

samson0v:
@devaskim we decided to do it this way because these methods will work only if we receive JSON data, so if we receive raw data this won't work.

thingsboard_gateway/connectors/mqtt/mqtt_converter_utils.py
devaskim: Done as discussed

devaskim:
@samson0v Done

Issue #942, opened 2022-10-08
[BUG] tb-gateway parses the received message as “INF“.


zlwjzf:
**Describe the bug**
When I access an MQTT protocol device and receive a message such as "42646174611E010000", the tb-gateway will parse the message as "INF", resulting in my custom "MqttCustomConverter.py" cannot parse the message content accurately

zlwjzf:
I also encountered messages that were parsed into data of type float, but I forgot to save the exact message contents. Please help to provide solutions, thank you!

devaskim:
Is "42646174611E010000" string or number on the broker side? And is this data wrapped in JSON object, e.g. `{ "my_variable": "42646174611E010000" }`?

zlwjzf:
@devaskim Hi, thanks for your attention
1、"42646174611E010000" is a string on the broker side
2、This data is not included in json

devaskim:
As I understand [decoding algorithm](https://github.com/thingsboard/thingsboard-gateway/blob/master/thingsboard_gateway/tb_utility/tb_utility.py#L31) for incoming mqtt data there is no quick fix for your case. 
MQTT Connector tries to interpret data as json-formatted and your string hits [this branch](https://github.com/thingsboard/thingsboard-gateway/blob/master/thingsboard_gateway/tb_utility/tb_utility.py#L36) as number in string.

devaskim:
I think it makes sense to add new configuration parameter for MQTT connector called e.g. `processRawData` to pass data to converter without any encoding.

@imbeacon @samson0v What do you think about this?

samson0v:
Hi @zlwjzf, please, send your MQTT Connector config file.

PR #945, opened 2022-10-12
Added ability to process incoming data without any decoding.

devaskim:
Discussion: #942

### DIFF
--- a/thingsboard_gateway/config/mqtt.json
+++ b/thingsboard_gateway/config/mqtt.json
@@ -1,4 +1,5 @@
 {
+  "processRawData": false,
   "broker": {
     "name": "Default Local Broker",
     "host": "127.0.0.1",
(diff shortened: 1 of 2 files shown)


samson0v:
Hi @devaskim, after review, I think your changes make no sense because the connector has to be able to process any type of data with a specific config but not a whole data.

If you are interested in releasing this feature, please look at the socket converter and how it processes raw data and its configuration.

Nonetheless, we will add this feature to MQTT Connector as soon as possible.

devaskim:
I am not sure what understood your arguments. So my question is what are you going to implement in MQTTConnector to cover #942 case and how your implementation would differ from mine?

devaskim:
IMHO, the role of mqtt connector is to provide connection between mqtt broker/data converters and Thingsboard server. At the moment mqtt connector limits gateway's users in JSON formatted data. If it wasn't so, we couldn't see #942

samson0v:
@devaskim config section will look like this:
```
{
    "type": "raw",
    "key": "[:]",
    "value": "[0:2]"
}
```
and the same for `deviceName` and `deviceType` parameters.

devaskim:
@samson0v Thanks for your explanation and for spending on me your time. This will be great feature. 

But to switch between json and raw traffic any way the mqtt connector will need some configuration option called `raw` as in your example above or something other name. I just tried to introduce this option now instead of near or further future.

Close the PR if you don't need this changes. And good luck!

PR #979, opened 2022-11-04
Fixed grpcio version in requirements.txt

samson0v:


helsmore:
@samson0v It looks like the file you've changed, `requirements.txt` in the repo root directory, doesn't get used when building the Docker image (if I've understood the Docker build instructions in the TB docs correctly).

From https://thingsboard.io/docs/iot-gateway/install/docker-installation/#build-local-docker-image:
1. Copy `thingsboard_gateway/` directory to `docker/` directory.
2. Build the Docker image

It seems that because you only copy this single directory from the repo root the Docker build process will end up using `docker/requirements.txt` not `requirements.txt`, this would mean that the change you've just made would not apply to the Docker image build, is that not the case?

What do you think?

cc: @imbeacon

helsmore:
In any case it appears there are some significant differences between `requirements.txt` and `docker/requirements.txt`. It's not clear why there's two different files, you could just delete `docker/requirements.txt` from the repo and copy `requirements.txt` at the same time as the `thingsboard_gateway` directory (whilst preparing for Docker build) unless there's some reason that you need different packages installed when running inside the Docker container compared to other deployment methods.

helsmore:
@imbeacon @samson0v 

Can you have a look at my comments above please? I think either the fix you made changed the wrong file or the docs are misleading.

Many thanks.

samson0v:
@helsmore done.

Issue #986, opened 2022-11-10
[Feature request] MQTT connector: Provide option to send data only on change


devaskim:
**Describe the issue**
At least `Modbus` and `CAN` connectors already [has this options](https://thingsboard.io/docs/iot-gateway/config/modbus/) so there is no need to explain for what it is.

Maybe it would be easier to move existing implementation to separate gateway class and call its API in connectors. @imbeacon  @samson0v what do think about this idea?

devaskim:
@imbeacon @samson0v Thanks for your time!
Please look at [my comment to PR](https://github.com/thingsboard/thingsboard-gateway/pull/991#issuecomment-1313446558)

devaskim:
@imbeacon @samson0v 
I forgot to mention it would be top of the top if such filtering module (shared between all connectors) would persist latest values to database in case of existing `storage` section is enabled. 
I know it is not the same but TB server already has separate database table for latest values. So gateway could go this way too.

devaskim:
@imbeacon @samson0v Please, take a look at #1000 (I am lucky - 1000th PR)

samson0v:
@devaskim see comments #1000

PR #1000, opened 2022-11-17
[Core] Data duplicate detector feature.

devaskim:
Discussion #986

1. `Duplicate detection` is performed after connector has pushed new converted data to gateway 's core module. So `detector` doesn't depend on any connector specificity.
2. By default, `duplicate detection` is disabled.
3. Each time new converted data is received by gateway's core, `detector` checks on `sendDataOnlyOnChange` option in that data.
4. If the option is present and has `false` value, data is not checked on duplicates.
5. If the option is present and has `true` value, data is checked on duplicates.
6. If the option is absent, `detector` asks the connector which has pushed data if `duplicate detection`  is enabled at all **(*)** or for the device - the origin of data.

**(*)** In scope of this work **ONLY** MQTT connector is a bit improved to inform if `duplicate detection` is enabled for this connector. Other connectors will respond that `duplicate detection` is disabled for them. Thus users of other connector should either set `sendDataOnlyOnChange` to `true` in each converted data (in other words, tune theirs converters) or contribute something similar that was done for MQTT connector in this PR.


thingsboard_gateway/connectors/mqtt/json_mqtt_uplink_converter.py
```diff
@@ -13,10 +13,10 @@
 #     limitations under the License.
 
 from re import search
-from time import time
 
 from simplejson import dumps
 
+from thingsboard_gateway.gateway.constants import *
```
samson0v, line 19: It is better to use exactly what you will use (in your case `SEND_ON_CHANGE_PARAMETER`).
`from thingsboard_gateway.gateway.constants import SEND_ON_CHANGE_PARAMETER`

thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -146,6 +147,7 @@ def __init__(self, config_file=None):
         global log
         log = logging.getLogger('service')
         log.info("Gateway starting...")
+        self.__duplicate_detector = DuplicateDetector(self)
```
samson0v, line 150: I think that there is no reason to pass a link to the whole Gateway instance (in DuplicateDetector using only self.__available_connectors)

thingsboard_gateway/connectors/mqtt/mqtt_connector.py
```diff
@@ -22,6 +22,8 @@
 
 import simplejson
 
+from thingsboard_gateway.gateway.constants import *
```
samson0v, line 25: It is better to use exactly what you will use

thingsboard_gateway/gateway/duplicate_detector.py
```diff
@@ -0,0 +1,104 @@
...
+#  #
+#      Unless required by applicable law or agreed to in writing, software
+#      distributed under the License is distributed on an "AS IS" BASIS,
+#      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#      See the License for the specific language governing permissions and
+#      limitations under the License.
+
+from logging import getLogger
+
+from thingsboard_gateway.gateway.constants import *
```
samson0v, line 17: It is better to use exactly what you will use

thingsboard_gateway/gateway/duplicate_detector.py
```diff
@@ -0,0 +1,104 @@
...
+        if remaining_attributes_count > 0 or remaining_telemetry_count > 0:
+            log.warning("[%s] '%s' changed attributes %d from %d, changed telemetry %d from %d",
+                        connector_name, device_name,
+                        remaining_attributes_count, remaining_attributes_count + filtered_attributes_count,
+                        remaining_telemetry_count, remaining_telemetry_count + filtered_telemetry_count)
+            to_send[DEVICE_NAME_PARAMETER] = device_name
+            to_send[DEVICE_TYPE_PARAMETER] = new_data[DEVICE_TYPE_PARAMETER]
+            return to_send
+
+        log.warning("[%s] '%s' device data has not been changed", connector_name, device_name)
```
samson0v, line 80: It will be better to use `log.info(...)`

thingsboard_gateway/gateway/duplicate_detector.py
```diff
@@ -0,0 +1,104 @@
...
+                    ts_added = True
+                    remaining_telemetry_count += 1
+                else:
+                    filtered_telemetry_count += 1
+            if ts_added:
+                to_send[TELEMETRY_PARAMETER].append(
+                    {TELEMETRY_TIMESTAMP_PARAMETER: ts, TELEMETRY_VALUES_PARAMETER: ts_values} if ts else ts_values)
+
+        if remaining_attributes_count > 0 or remaining_telemetry_count > 0:
+            log.warning("[%s] '%s' changed attributes %d from %d, changed telemetry %d from %d",
```
samson0v, line 72: It will be better to use `log.debug(...)`

thingsboard_gateway/gateway/duplicate_detector.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class DuplicateDetector:
+    def __init__(self, connectors_storage):
+        self._connectors_data = {}
+        self._connectors_storage = connectors_storage
+
+    def persist_latest_values(self):
+        # TODO: periodically save latest values if persistent storage is enabled
+        pass
```
samson0v, line 29: raise NotImplementedError

thingsboard_gateway/gateway/duplicate_detector.py
devaskim: Sorry, copy/paste. Maybe `debug`?

samson0v: We get a lot of issues where users don't see if data sending to TB, so I think it is better to see why it is happening without turning on DEBUG level.

thingsboard_gateway/gateway/tb_gateway_service.py
devaskim: ok

thingsboard_gateway/gateway/duplicate_detector.py
devaskim: ok

thingsboard_gateway/gateway/tb_gateway_service.py
devaskim: Well, my way is encapsulated and strongly ensured that duplicate detector get the latest updated list of connector after possible remote updates and some other features what may be added in the future. Agree?

thingsboard_gateway/connectors/mqtt/json_mqtt_uplink_converter.py
devaskim: ok

thingsboard_gateway/gateway/tb_gateway_service.py
samson0v: Yes, but self.__available_connectors is a dict that is a mutable object, so I think it must work too.

thingsboard_gateway/gateway/duplicate_detector.py
devaskim: ok

thingsboard_gateway/gateway/tb_gateway_service.py
devaskim: Done. Please, review again whole PR

samson0v:
@devaskim, I will test your changes

devaskim:
@samson0v what about [previous PR](https://github.com/thingsboard/thingsboard-gateway/pull/991#issuecomment-1318700131) for `sendDataOnlyOnChange`? Will it be reverted?

devaskim:
@samson0v Ok, I just want to ensure that this item is on tracking.

PR #1190, opened 2023-10-07
Fixed and improved default JSON MQTT uplink converter.

devaskim:
1. Add support of spaced keys.
2. Use `type` configuration option to convert incoming data to another data type.


devaskim:
@imbeacon @samson0v Do you need this PR or can I close it?

PR #1200, opened 2023-10-23
Meraki Sensors Timeseries 

lsenecaux:
Hi,
I have a problem with meraki sensors.
In fact, when their telemetries are sent back to the gateway, timestamps are not kept because they are provided as defined in RFC3339 (aka. 2023-10-23T15:00:00Z).
I therefore lose the history of these metrics.
Other types of timestamp format should be taken into account.

### DIFF
--- a/thingsboard_gateway/gateway/tb_gateway_service.py
+++ b/thingsboard_gateway/gateway/tb_gateway_service.py
@@ -12,6 +12,7 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
+import datetime
 import logging
 import logging.config
 import logging.handlers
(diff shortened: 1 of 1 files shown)


imbeacon:
Hi @lsenecaux,

Thanks for your interest in ThingsBoard IoT gateway. We understand your idea, added in this PR. But it covers only one corner case, in this case we should have some generic solution for all converters to allow users provide their date formats and give ability to dynamically configure that. 

Please add changes to make it generic, you can use "dateFormat" parameter in converter configuration, for example.
Or we can add it in one of next releases as a separated feature.

thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -984,8 +985,14 @@ def __convert_telemetry_to_ts(data):
                 if isinstance(item['ts'], int):
                     telemetry_with_ts.append({"ts": item["ts"], "values": {**item["values"]}})
                 else:
-                    log.warning('Data has invalid TS (timestamp) format! Using generated TS instead.')
-                    telemetry_with_ts.append({"ts": int(time() * 1000), "values": {**item["values"]}})
+                    try:
```
imbeacon, line 988: Probably it will be better to add a parameter for dateFormat to converter and resolve this point in converters, because it may be configurable by user. Also, this module is responsible about final data formatting and sending it to TB, it is not designed to use configuration for data formatting or anything else.

thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -984,8 +985,14 @@ def __convert_telemetry_to_ts(data):
...
+                        dt_obj = datetime.datetime.strptime(item["ts"], '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=datetime.timezone.utc)
```
imbeacon, line 989: The data format may be different, so it will be better to make it configurable.

thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -984,8 +985,14 @@ def __convert_telemetry_to_ts(data):
...
+                        timestamp = int(dt_obj.timestamp() * 1000)
+                        telemetry_with_ts.append({"ts": timestamp, "values": {**item["values"]}})
+                        log.warning('Data has string TS (timestamp) format! Convert to integer.')
+                    except:
```
imbeacon, line 993: After adding a dataFormat as a separated parameter - we can check is it present and if yes - use converting for incoming ts.

lsenecaux:
Indeed, I agree with you to add a user parameter to take into account the date format coming from the sensor

PR #1236, opened 2023-11-21
BACNet: Dont remove name prop from device

elgutierrez:


elgutierrez:
My bad, I thought that I was targeting my fork

PR #1270, opened 2023-12-27
opcua: Fixes issues with computing "information" based on path

wilterdinkrobert:
### Bugfix

Timeseries with nodeid paths containing `\\` in the string part failed with `re.error bad escape \... at position ...`. The `fullmatch(path, config)` would compile the `path` part as a regex, but cannot be assumed to be correctly escaped (hence the `re.error`). Upon closer inspection the rather complicated code can be removed completely by storing the `information` with the subscription when it is first made. Functionality from configuration pov should be unchanged/equivalent.

Test configuration:
```
        "timeseries": [
          {
            "key": "MachineRt_Error_Status",
            "path": "${ns=2;s=MachineRt\\Error\\Status}"
          },{
            "key": "MachineRt_Cycle_Status",
            "path": "${ns=2;s=MachineRt\\Cycle\\Status}"
          }]
```

Tested with several small OPCUA servers.

### DIFF
--- a/tests/connectors/opcua/test_opcua_connector.py
+++ b/tests/connectors/opcua/test_opcua_connector.py
@@ -1,0 +1,34 @@
+import logging
+import unittest
+from unittest import mock
+
+from opcua import Node
+from opcua.ua import FourByteNodeId
+
+from thingsboard_gateway.connectors.opcua.opcua_connector import OpcUaConnector
+
+logging.basicConfig(level=logging.ERROR,
+                    format='%(asctime)s - %(levelname)s - %(module)s - %(lineno)d - %(message)s',
+                    datefmt='%Y-%m-%d %H:%M:%S')
+log = logging.getLogger("root")
+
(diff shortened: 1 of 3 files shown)


imbeacon:
Hi @wilterdinkrobert,
Thank you for your interest and participating in the project life. I have checked the difference and could you confirm that configuration like the following one will still work correctly? Or the same path, but from the Root object on the server?

```
        "timeseries": [
          {
            "key": "MachineRt_Error_Status",
            "path": "${MachineRt\\.Error\\.Status}"
          },{
            "key": "MachineRt_Cycle_Status",
            "path": "${MachineRt\\.Cycle\\.Status}"
          }]
```

wilterdinkrobert:
Yes, I did test the partial browse_path. I've now also tested with the full browse_path (from Root) and encountered some bugs/inconsistencies (not related to the previous commits). Could you please confirm that the tests in 8f67463efde7ab3e84d8f0a3e3f951ebc16d80f9 match the intended behavior?

imbeacon:
Yes, looks good.

PR #1289, opened 2024-01-30
Added blackbox tests for Modbus RPC

samson0v:


tests/blackbox/connectors/modbus/test_modbus_rpc.py
```diff
@@ -0,0 +1,447 @@
...
+        username = GatewayDeviceUtil.DEFAULT_USERNAME
+        password = GatewayDeviceUtil.DEFAULT_PASSWORD
+
+        with RestClientCE(url) as cls.client:
+            cls.client.login(username, password)
+
+            cls.gateway = cls.client.get_tenant_devices(10, 0, text_search='Gateway').data[0]
+            assert cls.gateway is not None
+
+            while not cls.is_gateway_connected():
```
imbeacon, line 40: Please, add a timeout, because it may stuck the build machine.

tests/blackbox/connectors/modbus/test_modbus_rpc.py
```diff
@@ -0,0 +1,447 @@
...
+        except IndexError:
+            return False
+
+    @classmethod
+    def load_configuration(cls, config_file_path):
+        with open(config_file_path, 'r', encoding="UTF-8") as config:
+            config = load(config)
+        return config
+
+    def change_connector_configuration(self, config_file_path):
```
imbeacon: This method may be useful in other tests, could we move it to the base class for blackbox tests?

tests/blackbox/connectors/modbus/test_modbus_rpc.py
```diff
@@ -0,0 +1,447 @@
...
+            client.write_register(28, 10, 1)
+        except ConnectionException:
+            # will call pymodbus.exceptions.ConnectionException because of restarting the server
+            pass
+
+        client.close()
+        sleep(2)
+
+    @classmethod
+    def is_gateway_connected(cls):
```
imbeacon, line 70: It will be better to check the server attribute "lastConnectedTime" with "active" attribute, because active attribute is not a signal that the device is connected.

tests/blackbox/connectors/modbus/test_modbus_attributes_updates.py
```diff
@@ -46,6 +48,21 @@ def setUpClass(cls) -> None:
...
             assert cls.device is not None
 
+    @classmethod
+    def tearDownClass(cls):
+        super(ModbusAttributesUpdatesTest, cls).tearDownClass()
+
+        client = ModbusClient.ModbusTcpClient('modbus-server', port=5021)
+        client.connect()
+        try:
+            client.write_register(28, 10, 1)
```
imbeacon, line 63: Add comment what do we do here, because we won't remember it in a month.

PR #1359, opened 2024-03-29
bugfix(tb_gateway_service):fix bug that cannot return when RPC remote…

want-eat-meat:
when do RPC remote shell,it may cause cannot return


imbeacon:
Hi @want-eat-meat,

Thank you for your participating in project life, could you add logging instead of pass? It may be not completely clear why it fails silently. I think debug level will be enough.

PR #1391, opened 2024-05-09
Update tb_gateway_remote_configurator.py

ultimate-smile:
fix bug 1389 - Enabling or disabling the linker through platform gateway configuration causes devices under other linkers to be unable to communicate normally.


ultimate-smile:
bug fix 1389

PR #1468, opened 2024-07-18
Added collection of additional statistics

samson0v:


imbeacon:
Let’s send the gateway statistics every 10 minutes by default, instead of 1 hour.

Issue #1494, opened 2024-08-11
[BUG] Connector OPC-UA ASYNCIO creates too many sessions (BadTooManySessions) and do not release them when closing or disabling connector


ashdam:
**Describe the bug**

We have a compulsory "max number of sessions" within OPC-UA in our company which is 10 and a timeout of 20 minutes by default.
Give this restriction, we get "BadTooManySessions" many like this 
![image](https://github.com/user-attachments/assets/88d8b2fb-ad18-4b68-8133-57a9bc30c5cd)
after investigating we noticed that OPC-UA ASYNCIO creates sessions but do not close them in the following reproducible scenarios:

Starting point: 

2 connectors "disabled" from last session (where we couldn't work any more due to "BadTooManySessions")
![image](https://github.com/user-attachments/assets/02b034a6-c12c-48f5-9483-b7eb06c10ed6)

Scenarios/ not expected behaviours:

1) Once gateway is started:

- if TB edge PE is not available (down) ==> connections is not established which is OK.
- if TB edge PE is available (up) ==> it creates a new session, the 4th, even when connector is disabled. We started UAexpert with 3. This is the output of logs. We guess its because there were some pending data to be sent but its shouldnt anyways because connector is disabled in this case

Expected: Do not connect if connector is disabled.

```
2024-08-11 09:03:28 - |ERROR| - [tb_logger.py] - tb_logger - exception - 133 - Cannot put converted data!
Traceback (most recent call last):
  File "/home/admin/tb-iot-gateway-latest/thingsboard-gateway/venv/lib/python3.12/site-packages/thingsboard_gateway-3.5.1-py3.12.egg/thingsboard_gateway/gateway/tb_gateway_service.py", line 992, in send_to_storage
    filtered_data = self.__duplicate_detector.filter_data(connector_name, data)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'TBGatewayService' object has no attribute '_TBGatewayService__duplicate_detector'. Did you mean: '_TBGatewayService__device_filter'?
2024-08-11 09:03:28 - |DEBUG| - [opcua_uplink_converter.py] - opcua_uplink_converter - convert - 112 - Converted data: {'deviceName': 'MB107_UDC1_A0000KAO8B', 'deviceType': 'default', 'attributes': [ <DELETED_DATA_TO SHORTEN_LOG> ], 'telemetry': [<DELETED_DATA_TO SHORTEN_LOG>]}
[...]

PR #1495, opened 2024-08-11
Disconnects from OPC-UA server when connector is stopped so session is released

ashdam:
Disconnects from OPC-UA server when connector is stopped so session is released


ashdam (commented):
Sessions are not released after disabling the connector

More info here: https://github.com/thingsboard/thingsboard-gateway/issues/1494

PR #1497, opened 2024-08-12
Release-session-when-closing-OPC-UA-Connector

ashdam:
OPC-UA Connector now disconnects from OPC-UA server when closed so it releases the session


ashdam:
It's my first pull request :) could not solve the problem with "Gonzalo Moran" which is another email i used by mistake

PR #1499, opened 2024-08-14
OPC-UA release sessions when close

ashdam:
Ensures client disconnects and therefore the session when closing connector

### DIFF
--- a/thingsboard_gateway/connectors/opcua/opcua_connector.py
+++ b/thingsboard_gateway/connectors/opcua/opcua_connector.py
@@ -118,6 +118,7 @@ class OpcUaConnector(Connector, Thread):
         self.__log.info("Stopping OPC-UA Connector")
 
         asyncio.run_coroutine_threadsafe(self.__cancel_all_tasks(), self.__loop)
+        asyncio.run(self.__disconnect())
 
         start_time = monotonic()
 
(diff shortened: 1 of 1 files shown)


ashdam:
This would solve this bug 
https://github.com/thingsboard/thingsboard-gateway/issues/1494

thingsboard_gateway/connectors/opcua/opcua_connector.py
```diff
@@ -118,6 +118,7 @@ def close(self):
         self.__log.info("Stopping OPC-UA Connector")
 
         asyncio.run_coroutine_threadsafe(self.__cancel_all_tasks(), self.__loop)
+        asyncio.run(self.__disconnect())
```
samson0v, line 121: We need to run it before cancel_all_tasks and using the same run_coroutine_threadsafe function, so the code will look like:
```python
asyncio.run_coroutine_threadsafe(self.__disconnect(), self.__loop)
asyncio.run_coroutine_threadsafe(self.__cancel_all_tasks(), self.__loop)
```

thingsboard_gateway/connectors/opcua/opcua_connector.py
```diff
@@ -134,6 +135,13 @@ async def __cancel_all_tasks(self):
         for task in asyncio.all_tasks(self.__loop):
             task.cancel()
 
+    async def __disconnect(self):
+        try:
+            await self.__client.disconnect()
+            self.__log.info('%s has been disconnected from OPC-UA Server.', self.get_name())
+        except:
```
samson0v, line 142: We need to have the error to user see why exactly this situation happened:
```python
except Exception as e:
```

thingsboard_gateway/connectors/opcua/opcua_connector.py
```diff
@@ -134,6 +135,13 @@ async def __cancel_all_tasks(self):
...
+             self.__log.warning(f'{self.get_name()} could not be disconnected from OPC-UA Server.')
```
samson0v, line 143: It is not the best practice to use f'...', so I'd like to suggest using the following code here:
```python
self.__log.error('%s could not be disconnected from OPC-UA Server: %s', self.name, e)
```

PR #1552, opened 2024-10-04
custom rpc execution

khalilKian:
Hello,

I’d like to contribute to the custom RPC importing functionality for the gateway.

For a long time, I've wanted to enhance the gateway by adding more RPC methods that can be used for local network hardware communication or third-party integration, without exposing the rpcRemoteShell feature. This will provide a secure and controlled way to manage the gateway remotely.

Additionally, I’ve included some minor fixes related to the generation of the .deb package.


khalilKian:
guide:
 you need to create a file inside /etc/thingsboard-gateway/rpc/{filename}.py ( optional ) 
 
 filename.py 
 def __rpc_hello(self,*args):
    log.debug("hello RPC request received with arguments %s", args)
    return {"code": 200, "resp": "world", "action": self.version , "args" : arg[0]}

PR #1572, opened 2024-10-28
Fixed returning response from restart RPC

samson0v:


thingsboard_gateway/gateway/tb_gateway_service.py
```diff
@@ -1571,7 +1571,7 @@ def __rpc_gateway_processing(self, request_id, content):
             log.info("Remote shell is disabled.")
             method_function = self.__gateway_rpc_methods.get(method_to_call)
         if method_function is None and method_to_call in self.__rpc_scheduled_methods_functions:
-            seconds_to_restart = arguments * 1000 if arguments and arguments != '{}' else 0
+            seconds_to_restart = arguments * 1000 if arguments and arguments != '{}' else 1000
```
imbeacon, line 1574: probably it will be better to use max function as well, to avoid values less then 1 second, like `seconds_to_restart = max(seconds_to_restart, 1000)`

imbeacon (changes requested):

imbeacon (approved):

Issue #1672, opened 2025-02-06
[HELP] MODBUS Server don't do anything, don't attach to port


palandri:
Hello everyone.

**Describe the issue**
When configuring the MODBUS connector as Server (TCP or Serial), literally nothing happens. tb-gateway don't attach to port, no log message, literally nothing.

If I enable `sendDataToThingsBoard`, it open/attach to designated port, but start polling, working as a Client, but do not listen on the interface, only sending requests. This behaviour happens both in TCP or Serial mode. In serial, I can see traffic on the bus (the requests being sent by the Server).

I don't think this behaviour is correct... in my understanding, a Modbus Server should listen and respond to requests, not initiate requests to another device (which is the role of the client).

Also, to corroborate with the information shown above, if I ran a Server/Slave simulator on the bus, it answer correctly to the requests, showing (in my understanding) that the connector is indeed working as a Client/Master.

OTOH, running a Client/Master simulator on the bus, it keeps timing out, or gives a checksum error (I suppose when the requests conflict between the Simulator and the Server/Slave Connector (which is bad behaving as a Client/Master).

Before enabling `sendDataToThingsBoard`
![Image](https://github.com/user-attachments/assets/4a028205-9291-449b-8508-7766e14e8315)

After enabling `sendDataToThingsBoard` (no restart, only enabling)
![Image](https://github.com/user-attachments/assets/d1cbd3f8-b8be-482a-976d-5f238734dded)

Slave simulator answering the requests of the "Server"
![Image](https://github.com/user-attachments/assets/14c2cfc3-cd3e-4d64-82ea-d0c35d2e6382)

Master simulator timing out/giving error
![Image](https://github.com/user-attachments/assets/9e18fa1d-66ec-4ad1-9a16-28f1438ffb5e)

**Configuration** (Attach your configuration file)
> {
  "thingsboard": {
    "host": "redacted",
    "port": redacted,
    "remoteShell": false,
    "remoteConfiguration": true,
    "latencyDebugMode": false,
    "statistics": {
      "enable": true,
      "statsSendPeriodInSeconds": 3600
    },
    "deviceFiltering": {
      "enable": false,
      "filterFile": "list.json"
    },
    "maxPayloadSizeBytes": 8196,
    "minPackSendDelayMS": 50,
    "minPackSizeToSend": 500,
    "checkConnectorsConfigurationInSeconds": 60,
    "handleDeviceRenaming": true,
    "security": {
      "type": "accessToken",
[...]

PR #1675, opened 2025-02-11
Fix modbus serial

palandri:
This PR reflects my changes to the modbus connector to work in a serial environment.
I tried to maintain the overall architecture of the current code, but changed the logic to work with ModbusSparseDataBlocks.

Please, if possible, give some feedback. I'm running this implementantion atm in my local test setup.

Should close [#1672](https://github.com/thingsboard/thingsboard-gateway/issues/1672).

### DIFF
--- a/thingsboard_gateway/connectors/modbus/server.py
+++ b/thingsboard_gateway/connectors/modbus/server.py
@@ -229,7 +229,7 @@ class Server(Thread):
         blocks = {}
         if (config.get('values') is None) or (not len(config.get('values'))):
             self.__log.error("No values to read from device %s", config.get(DEVICE_NAME_PARAMETER, 'Modbus Slave'))
-            return
+            return None, None
 
         for (key, value) in config.get('values').items():
             values = {}


palandri:
Marked as a draft, waiting for feedback.

thingsboard_gateway/connectors/modbus/server.py
```diff
@@ -123,7 +172,8 @@ def get_slave_config_format(self):
             **self.__config,
             'deviceName': self.device_name,
             'deviceType': self.device_type,
-            'pollPeriod': self.poll_period
+            'pollPeriod': self.poll_period,
+            'no_master': True   # shows that this slave don't have a master
```
imbeacon, line 176: It looks like it always be 'no_master': True. Because you at first unpack config and then overwrite no_master it with True.

Also are you sure that we actually need no_master parameter in general? Because if connector can reads data from data blocks - master will be no longer required

palandri: Hello, @imbeacon, thanks for your feedback.

Yes, I think you are right in this regard. I tried to modify as little as possible of the Slave class, to not break stuff, so I created the 'no_master' parameter to keep the current implementantion of the Slave class working in polling mode. But it isnt a parameter of the Server class, but Slave class, initialized [here](https://github.com/palandri/thingsboard-gateway/blob/3283c0bb375ce5f414cb100e41e0aca01ca60145/thingsboard_gateway/connectors/modbus/slave.py#L61). When `get_slave_config_format` is called, it creates a Slave object (inside Server context), with `no_master = True`. If the Slave object is called outside of Server context, `no_master` do not exists, so it defaults to False.

With the gateway working as a Client, it makes sense to poll Servers (which is triggered by the Slave-class object).
On the other hand, with the gateway working as a Server, polling doesn't make sense anymore (as the Server need to react to when data is sent to it).

So, 'no_master' simply control if Slave' internal timer is to be activated or not.

I think a better implementation of it would be to change Slave class to initialize the timer based on `poll_period` parameter (for example, if `poll_period = 0`, do not poll).

palandri:
Hello, @imbeacon, thanks for your feedback.

> If data on the server was modified by external master, will it be updated in data block?

Yes, it will. When a master writes to the server, it calls `ModbusSparseDataBlock.setValues` internally, overridden [here](https://github.com/palandri/thingsboard-gateway/blob/3283c0bb375ce5f414cb100e41e0aca01ca60145/thingsboard_gateway/connectors/modbus/server.py#L62), triggering AsyncModbusConnector callback, which finally enter the older dataflow (tested and proven) and send data upstream.

> If we modify the data after server starting - will external master receive new data or still retrieve the old?

Data is correctly written/retrieved in gateway scope, I did not test in Thingsboard scope. For example, if another entity (on Thingsboard context) change an attribute of the corresponding device in the gateway, will a Client reading the Server fetch this new data?

I'll test and report it.

imbeacon:
Looks like it needs to send uplink right after server started to keep value on TB updated.

According to shared attributes updates and RPC - I think we need to adjust shared attributes updates/RPC logic for server in Modbus Connector. You can do this or we will do this after merge, it is not a problem. 

Thank you, we appreciate your contribution.

palandri:
> Looks like it needs to send uplink right after server started to keep value on TB updated.
According to shared attributes updates and RPC - I think we need to adjust shared attributes updates/RPC logic for server in Modbus Connector. You can do this or we will do this after merge, it is not a problem.

Oh, I see. To be honest, I did not test shared attributes updates/RPC, only attributes and timeseries. I'll test it and propose changes, no problem.

> Turns out gateway maps attributes as Client-side (which is correct), so you can't modify attributes in Thingsboard (which makes sense). So, attributes/telemetry do not propagate downstream, only upstream.

The reported behaviour is specific to attributes/telemetry, not shared attributes. So, as I'm understanding, this behaviour is correct. Could you please confirm, @imbeacon?

Also, about the overall implementation and/or style, could you share some constructive feedback? I'm always looking for opportunities to grow as a developer.

imbeacon:
Hi @palandri,

First of all, I appreciate your contribution and the effort you’ve put into improving the ModbusConnector!

> The reported behavior is specific to attributes/telemetry, not shared attributes. So, as I’m understanding, this behavior is correct. Could you please confirm, @imbeacon?

If you’re referring to updating values right after the server starts, then yes, that behavior is correct.

Regarding the overall implementation, I agree with your approach, but I believe we can simplify things further. Specifically, we can eliminate the need for the no_master parameter when the gateway operates as a server. Since we can now read the actual state of sparse blocks and update them without requiring a client, keeping the client as-is may lead to issues—especially with shared attribute updates or incoming RPCs, where an inability to connect to the serial port could arise.

A more effective approach would be to remove the client when the gateway operates as a server across all communication types (Serial/TCP/UDP). This way, the Slave class wouldn’t need any modifications, as it wouldn’t be used in this scenario. Instead, all necessary adjustments would be contained within the Server class and ModbusConnector.

Does this approach make sense to you? While it’s slightly more complex than the initial idea, it provides a cleaner abstraction and avoids unnecessary client creation, making the implementation more efficient in the long run.

Looking forward to your thoughts!

PR #1718, opened 2025-04-04
Add useReceivedTs option for consistent timestamp handling

rroemhild:
# Add option to use received timestamp for all timeseries in MQTT Connector

## Description
Added a new configuration parameter `useReceivedTs` that allows using the same received timestamp for all timeseries data points within a single message. This helps maintain consistency in timestamp values across related measurements that are received together.

The constant `USE_RECEIVED_TS_PARAMETER = "useReceivedTs"` was added to support this feature.

## Changes
- Added new constant `USE_RECEIVED_TS_PARAMETER = "useReceivedTs"` in constants.py
- When enabled, the received timestamp will be used for all timeseries data points in a message instead of individual timestamps
- This helps maintain time consistency across related measurements received in a single message

## Benefits
- Ensures all related measurements within a message have the same timestamp
- Prevents time skew between measurements that logically occurred together
- Makes it easier to correlate related datapoints in time-series analysis
- Provides more accurate representation of when data was actually received

The changes are backwards compatible since the `useReceivedTs` parameter is optional.

### DIFF
--- a/thingsboard_gateway/connectors/mqtt/json_mqtt_uplink_converter.py
+++ b/thingsboard_gateway/connectors/mqtt/json_mqtt_uplink_converter.py
@@ -19,7 +19,7 @@ from simplejson import dumps
 
 from thingsboard_gateway.connectors.mqtt.mqtt_uplink_converter import MqttUplinkConverter
 from thingsboard_gateway.gateway.constants import REPORT_STRATEGY_PARAMETER, \
-    RECEIVED_TS_PARAMETER, CONVERTED_TS_PARAMETER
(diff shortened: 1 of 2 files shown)


imbeacon:
Hi @rroemhild,

Thank you for your interest in ThingsBoard IoT Gateway and contribution.
It is a good functionality, but could you move USE_RECEIVED_TS_PARAMETER = "useReceivedTs" to converter itself? Because gateway.constants it is a global scope and adding to it may be correct in case of adding this functionality to all converters. In current case it will be enough to add this to `json_mqtt_uplink_converter.py` file or you may create a new file in connectors/mqtt folder with name constants.py and use it.

PR #1740, opened 2025-04-17
Added debugpy

samson0v:


setup.py
```diff
@@ -87,7 +87,8 @@
         'tb-mqtt-client>=1.13.5',
         'packaging==23.1',
         'service-identity',
-        'psutil'
+        'psutil',
```
imbeacon, line 90: You are installing debugpy in runtime, it is not required in setup.py and requirements.txt

PR #1750, opened 2025-04-24
feature: added more logs, improve existing logs messages

Andry925:
Added more logs for ftp connector, improved log messages for existing.

### DIFF
--- a/thingsboard_gateway/connectors/ftp/ftp_connector.py
+++ b/thingsboard_gateway/connectors/ftp/ftp_connector.py
@@ -93,7 +93,7 @@ class FTPConnector(Connector, Thread):
             )
             for obj in self.config['paths']
         ]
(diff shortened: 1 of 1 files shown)


thingsboard_gateway/connectors/ftp/ftp_connector.py
```diff
@@ -115,16 +116,20 @@ def run(self):
                             if self.__stopped:
                                 break
         except Exception as e:
-            self.__log.exception(e)
+            self.__log.error("Unexpected exception in loop for %s %d with error %s" % (self.host, self.port, str(e)))
```
imbeacon, line 119: Please, use the following format of calling logs method, it will avoid unnecessary formatting:

"Unexpected exception in loop for %s %d with error %r", self.host, self.port, e)

PR #1755, opened 2025-04-28
feature:add functionality to parse timestamp from different connectors

Andry925:
Added functionality to parse timestamp field from request, rest, mqtt connectors.


thingsboard_gateway/tb_utility/tb_utility.py
```diff
@@ -26,6 +27,7 @@
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.x509.oid import NameOID
+from distutils.util import strtobool
```
imbeacon, line 30: It is not necessary to use this library, also it provokes issue with python 3.12+, because it is no longer built-in, please remove this change and use removed by you str_to_bool method instead.

PR #1757, opened 2025-04-28
Added BACnet device objects autodiscovery

samson0v:


thingsboard_gateway/connectors/bacnet/application.py
```diff
@@ -31,10 +36,90 @@ def __init__(self, device_object_config: DeviceObjectConfig, indication_callback
...
+    async def get_object_identifiers(
+        self, device_address: Address, device_identifier: ObjectIdentifier
+    ) -> List[ObjectIdentifier]:
+        try:
+            object_list = await self.read_property(
+                device_address, device_identifier, "objectList"
+            )
+            return object_list
+        except AbortPDU as err:
+            if err.apduAbortRejectReason != AbortReason.segmentationNotSupported:
```
imbeacon, line 56: What if error is AbortReason.segmentationNotSupported?

In IAm response usually device returns supported segmentation and other parameters, this can be easily discovered using Wireshark, probably we should base on them for interactions with device, because current implementation may lead to loosing data from part of devices in the same network.

thingsboard_gateway/connectors/bacnet/application.py
```diff
@@ -31,10 +36,90 @@ def __init__(self, device_object_config: DeviceObjectConfig, indication_callback
...
-        self.__indication_callback(apdu)
+        await self.__indication_callback(apdu)
 
     async def do_who_is(self, device_address):
         devices = await self.who_is(address=Address(device_address),
                                     timeout=self.__device_object_config.device_discovery_timeout)
         if len(devices):
             return devices[0]
+
+    async def get_object_identifiers(
```
imbeacon, line 47: Could we move methods get_object_identifiers and get_device_objects to Device class? it looks more convenient to interact with it, don't you think so?

PR #1765, opened 2025-05-05
fix:fix rpc methods in ftp connector

Andry925:
Fixed ftp connector to correctly process all required rpc methods

### DIFF
--- a/thingsboard_gateway/connectors/ftp/ftp_connector.py
+++ b/thingsboard_gateway/connectors/ftp/ftp_connector.py
@@ -18,7 +18,6 @@ from ftplib import FTP, FTP_TLS
 from queue import Queue
 from random import choice
 from re import fullmatch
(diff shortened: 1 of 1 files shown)


thingsboard_gateway/connectors/ftp/ftp_connector.py
```diff
@@ -415,6 +421,7 @@ def __process_rpc(self, method, value_expression):
                     ftp.storbinary('STOR ' + arr[0], io_stream)
                     io_stream.close()
                     success_sent = True
+                    converted_data = f'The value is written to {arr[0]}' if len(arr) < 80 else True
```
imbeacon, line 424: Please change this, to print logs, with some string, but RPC should return something like {"result": value}, you can use OPC-UA connector as a reference in this case.

PR #1774, opened 2025-05-09
change log message from info to debug for rpc_gateway_processing

Andry925:


imbeacon:
These RPC are rare and these logs may be helpful, during some investigations that may appear , even when user don't have debug enabled, they are not spam or something like this. So there is no reason to change current behavior.

PR #1775, opened 2025-05-09
remove deprecated functions to process configs from mqtt, opcua and t…

Andry925:
…b_gateway_service


imbeacon:
Please remove default implementation for connector of this method as well.

PR #1782, opened 2025-05-15
BACnet improvements

samson0v:


thingsboard_gateway/connectors/bacnet/device.py
```diff
@@ -72,6 +74,19 @@ def __str__(self):
...
+        return self.__stopped
+
+    @property
+    def reading_time(self):
+        return self.__reading_time
+
+    @reading_time.setter
+    def reading_time(self, new_reading_time):
+        self.__reading_time = new_reading_time
+        self.__poll_period += new_reading_time
```
imbeacon, line 88: It looks wrong to increase poll_period on each assign. If reading time  more than poll period from config - we should assign reading time to poll period for next request.

thingsboard_gateway/connectors/bacnet/entities/bacnet_device_details.py
```diff
@@ -38,3 +54,9 @@ def as_dict(self):
             "vendorId": self.__vendor_id,
             "objectName": self.__object_name
         }
+
+    def is_segmentation_supported(self):
+        return self.__segmentation in (Segmentation.segmentedBoth, Segmentation.segmentedTransmit)
+
+    def get_max_apdu_count(self) -> int:
+        return int(self.__max_apdu_length / 26)  # TODO: move 26 to constant
```
imbeacon, line 62: ?

imbeacon (changes requested):

PR #1799, opened 2025-05-22
BACnet add object/device identifier filter

granaderainer:
add objectIdFilter to device config.
In case host key represents more then one ip (e.g. *, ...) it's now possible to filter the responding devices based on their device identifier.

objectIdFilter

In case host is set to one ip address and objectIdFilter is set and identifier and ip are not from same device, there is no config match.

Possible adaptions:
maybe its better to allow just one identifier for filtering (move from array > str) or use regex for filtering?

@imbeacon @samson0v what are your thoughts about this?

### DIFF
--- a/thingsboard_gateway/connectors/bacnet/device.py
+++ b/thingsboard_gateway/connectors/bacnet/device.py
@@ -132,11 +132,48 @@ class Device:
         apdu_address = apdu.pduSource.__str__()
 
         for device_config in devices_config:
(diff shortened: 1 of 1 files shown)


imbeacon:
Hi @granaderainer,

Thank you for your contribution, we will add related functionality in the following PR - https://github.com/thingsboard/thingsboard-gateway/pull/1859

There will be included different abilities to configure devices filtration and discovering by identifiers.

PR #1822, opened 2025-06-12
Added ability to limit database file sizes in SQLite storage

Andry925:
Implemented functionality to


tests/unit/service/test_storage.py
```diff
@@ -1,6 +1,8 @@
 from logging import getLogger
-from os import listdir, remove, removedirs
+from os import listdir, remove, removedirs, path
 from random import randint
+from shutil import rmtree
+from tempfile import mkdtemp
```
imbeacon, line 5: The gateway should automatically create required folders, you don't need to do this manually.

thingsboard_gateway/storage/sqlite/database.py
```diff
@@ -67,30 +67,34 @@ def __init__(self, config, processing_queue: Queue, logger, stopped: Event):
...
         try:
-            # Check if the old schema exists
             try:
+
                 result = self.db.execute_read(
                     "SELECT sql FROM sqlite_master WHERE type='table' AND name='messages';"
                 ).fetchone()
+
+            except Exception as e:
+                result = None
```
imbeacon, line 79: It may fail silently here.

thingsboard_gateway/storage/sqlite/sqlite_event_storage.py
```diff
@@ -1,113 +1,352 @@
...
     def event_pack_processing_done(self):
-        self.__log.trace("Batch processing done, processing time: %i milliseconds",
-                        int((monotonic() - self.__event_pack_processing_start) * 1000))
+        self.__log.trace(
+            "Batch done in %d ms",
+            int((monotonic() - self.__event_pack_processing_start) * 1000),
+        )
         if not self.stopped.is_set():
-            self.delete_data(self.delete_time_point)
+            try:
```
imbeacon, line 220: It should raise the error here, to handle it in the upper level.

thingsboard_gateway/storage/sqlite/storage_settings.py
```diff
@@ -12,10 +12,25 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
+from os import path
+
+
 class StorageSettings:
     def __init__(self, config):
         self.data_folder_path = config.get("data_file_path", "./")
```
imbeacon, line 20: Looks like at the moment, this parameter means data_file_path, not data_folder_path anymore. Please rename it.

thingsboard_gateway/storage/sqlite/sqlite_event_storage.py
```diff
@@ -1,113 +1,352 @@
...
-            sleep(.1)
+        self.__read_database.close_db()
+        self.__write_database.close_db()
         collect()
 
     def len(self):
-        qsize = self.write_queue.qsize()
-        stored_messages = self.db.get_stored_messages_count()
-        return qsize + stored_messages
+        return self.write_queue.qsize() + self.__read_database.get_stored_messages_count()
```
imbeacon, line 349: What about situation, when read_database was changed to the new one? or we have several databases? We need to return here amount of stored events, not just in current read database.

thingsboard_gateway/storage/sqlite/sqlite_event_storage.py
```diff
@@ -1,113 +1,352 @@
...
         super().__init__(config, logger, main_stop_event)
         self.__log = logger
         self.__log.info("Sqlite Storage initializing...")
         self.write_queue = Queue(-1)
         self.stopped = Event()
-        self.db = Database(config, self.write_queue, self.__log, stopped=self.stopped)
-        self.db.init_table()
-        self.__log.info("Sqlite storage initialized!")
-        self.delete_time_point = None
+        self.__config_copy = copy.deepcopy(config)
```
imbeacon, line 24: It will be better to work with storage settings object, instead of copying dict with config. You can adjust storage settings, for example make copy() method, which will return a new instance with the same values in attributes. And then adjust a new settings object with required data.

imbeacon (changes requested):

thingsboard_gateway/storage/sqlite/sqlite_event_storage.py
Andry925: Fixed

thingsboard_gateway/storage/sqlite/sqlite_event_storage.py
Andry925: Fixed

PR #1873, opened 2025-07-09
Bugfixs for Request connector JSON downlink converter

apachler:
Same bugs fixed as in REST JSON downlink converter

### DIFF
--- a/thingsboard_gateway/connectors/request/json_request_downlink_converter.py
+++ b/thingsboard_gateway/connectors/request/json_request_downlink_converter.py
@@ -12,7 +12,7 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
-import ast
+from ast import literal_eval
 from urllib.parse import quote
 
 from simplejson import dumps
@@ -66,7 +66,7 @@ class JsonRequestDownlinkConverter(RequestConverter):
             for (tag, value) in zip(data_tags, data_values):
                 result['data'] = result["data"].replace('${' + tag + '}', str(value))
 
-            result["data"] = dumps(ast.literal_eval(result["data"]))
+            result["data"] = dumps(literal_eval(result["data"]))
             return result
         except Exception as e:
             self.__log.exception(e)
--- a/thingsboard_gateway/connectors/request/json_request_downlink_converter.py
+++ b/thingsboard_gateway/connectors/request/json_request_downlink_converter.py
@@ -12,7 +12,10 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
+import ast
 from urllib.parse import quote
+
+from simplejson import dumps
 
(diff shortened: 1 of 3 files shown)


thingsboard_gateway/connectors/request/json_request_downlink_converter.py
```diff
@@ -12,8 +12,11 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
+import ast
```
imbeacon, line 15: Could you use `from ast import literal_eval`?

imbeacon:
Also, these changes may broke the following part of code in connector - `params['json'] = params.pop("data")`, so please check that these changes won't break processing outgoing requests with `"Content-Type": "application/json"` header.

apachler:
Ahhh, ok, the implementation of REST and Request connector is different for handling JSON downlink messages.

https://github.com/thingsboard/thingsboard-gateway/blob/68b5e3989d6a12206ec9eb5664e7a727e63395ea/thingsboard_gateway/connectors/request/request_connector.py#L287-L290

https://github.com/thingsboard/thingsboard-gateway/blob/68b5e3989d6a12206ec9eb5664e7a727e63395ea/thingsboard_gateway/connectors/rest/rest_connector.py#L425-L426


Would you like me to leave this difference untouched and make my changes work with Request connector?
Or do you want me to change the REST connector so it works like the Request connector and to fix my changes for both connectors?

imbeacon:
This part is similar in the both connectors, please change the both. You can check the type or convert to bytes right before sending request, if content type is not a json.

apachler:
> This part is similar in the both connectors, please change the both. You can check the type or convert to bytes right before sending request, if content type is not a json.

Sorry, I dont get the point here ... in my opinion it would be the best to change both connectors the following way:
- As given, the JSON converter returns a `str`. (should not be changed for backward compatibility)
- Both connectors should not move data to `json` parameter of `requests` because it expects a serializable Python object. Since we already serialize the Python object to `str` in the converter, this step would be redundant.
- Both connectors only encodes `str` to `utf-8` bytes. This would give the possibility to custom converters to return other data types that `requests` can handle.

What do you think?

apachler:
> > This part is similar in the both connectors, please change the both. You can check the type or convert to bytes right before sending request, if content type is not a json.
> 
> Sorry, I dont get the point here ... in my opinion it would be the best to change both connectors the following way:
> 
> * As given, the JSON converter returns a `str`. (should not be changed for backward compatibility)
> * Both connectors should not move data to `json` parameter of `requests` because it expects a serializable Python object. Since we already serialize the Python object to `str` in the converter, this step would be redundant.
> * Both connectors only encodes `str` to `utf-8` bytes. This would give the possibility to custom converters to return other data types that `requests` can handle.
> 
> What do you think?

@imbeacon any feedback please?

PR #1874, opened 2025-07-09
Added data unpack functionality in Request connector

apachler:
Unpack data if `dataUnpackKey` is defined in config
```json
{
  "mapping": [
    {
      "dataUnpackKey": "device"
    }
  ]
}
```

This allows to unpack JSON responses that have final data at a sub key
```json
{
   "device": [...]
}
```

### DIFF
--- a/thingsboard_gateway/config/request.json
+++ b/thingsboard_gateway/config/request.json
@@ -20,6 +20,7 @@
       "allowRedirects": true,
       "timeout": 0.5,
       "scanPeriod": 5,
+      "dataUnpackExpression": "",
       "converter": {
         "type": "json",
         "deviceNameJsonExpression": "SD8500",
@@ -54,6 +55,7 @@
       "allowRedirects": true,
       "timeout": 0.5,
       "scanPeriod": 100,
+      "dataUnpackExpression": "",
       "converter": {
         "type": "custom",
         "deviceNameJsonExpression": "SD8500",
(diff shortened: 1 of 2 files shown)


imbeacon:
What if the response has several levels?

Also, you can use JSONPath expressions to parse data from the responses. Like `${$.device.someKey}`

apachler:
Does the JSONPath solution work if the sub key is an array of objects where i want to process every item as separate device?

imbeacon:
At the moment, it cannot, but it can be done by changing the uplink converter, for example add configuration parameter `devicesArray`, that will contain JSONPath expression and in uplink converter retrieve these object from the response using TBUtility, and if it is a list - iterate over it and parse device names, types, and other data for these devices. 

What do you think?

apachler:
But why implementing iteration into the converter since the connector already handels lists at https://github.com/thingsboard/thingsboard-gateway/blob/a999d41ae79731f4a537d5ff3639b3d419fe53e8/thingsboard_gateway/connectors/request/request_connector.py#L330

So, maybe it is enough to add support for JSONPath expressions to my change?

imbeacon:
> But why implementing iteration into the converter since the connector already handels lists at
> 
> https://github.com/thingsboard/thingsboard-gateway/blob/a999d41ae79731f4a537d5ff3639b3d419fe53e8/thingsboard_gateway/connectors/request/request_connector.py#L330
> 
> So, maybe it is enough to add support for JSONPath expressions to my change?

Yes, it may be enough.

apachler:
Added support for JSONPath expressions ...

PR #1875, opened 2025-07-09
Fixed problem with attribute updates that contain JSON data

apachler:


imbeacon:
Please avoid modifications of lines, not related to your changes, it is confusing and not necessary.

apachler:
Oh sorry, this comes from my IDEs code formater ... going to undo it ...
Or do you have a code formatter in use? Maybe something like Black?

apachler:
@imbeacon reverted the code formatter changes...

imbeacon:
At the moment we don't use any special code formatter. Also I would ask you not to use force pushes, because it may cause conflicts.

PR #1876, opened 2025-07-09
Added support for sub requests in Request connector

apachler:
This feature adds the possibility to do sub requests based on the data retrieved by an initial request.

As example, the initial request offers `id` key where we want to do a sub request to another URL that needs the `id` from the initial request.

### intial request
URL: http://xyz/devices

Response:
```json
{
  "id": 1,
  "name": "Test"
}
```

### sub request
URL: http://xyz/devices/${id}/status

Response:
```json
{
  "status": "OK"
}
```

This gives the possibility to implement dynamic device data gathering based on the response of the initial request.

In combination with https://github.com/thingsboard/thingsboard-gateway/pull/1874 the initial response could also give a list of devices on a sub key that are sub-sequentially requested like described above.

The data retrieved by the sub request is merged with the data of the initial request.
It is also possible to define a processing function for the data retrieved by the sub request. This function can be used to manipulate the retrieved data before the merge is done.

### DIFF
--- a/thingsboard_gateway/config/request.json
+++ b/thingsboard_gateway/config/request.json
@@ -133,16 +133,16 @@
       "allowRedirects": true,
       "timeout": 3.0,
       "scanPeriod": 1800,
+      "subRequests": {
+        "cumPwr": {
+          "url": "/${id}/hist/values/ActiveEnergy/SUM13/3600?start=RELATIVE_-1HOUR&end=RELATIVE_-1HOUR&online=false&aggregate=false",
+          "processingFunction": "def process_data(data, key): return {key: float(data[\"values\"][0][\"avg\"])/1000.0}"
+        }
+      },
       "converter": {
         "type": "json",
         "deviceNameJsonExpression": "${id}-${type}",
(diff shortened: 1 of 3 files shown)


apachler:
I am going to undo the unneeded changes by the formater ...

apachler:
@imbeacon reverted the code formatter changes...

imbeacon:
This idea is interesting, I think it may be extended with some kind of "requestsChain", when in configuration we have an object with keys "0","1"... with configuration objects for request and additional fields like `inputs` and `outputs` and requestUrlExpression, requestBodyExpression... - in this case there will be ability to use as much requests as you want.

Also, Converters are designed to just convert data, they are not responsible about connection or data collection, please move this logic into connector itself.

apachler:
Ok, i have to think about the "requestChain" functionality but sure i can move the actual logic to the connector ...

imbeacon:
Please move it connector and I will merge this as initial implementation, requestChain may be in future improvements.

apachler:
Moved to connector...

PR #1878, opened 2025-07-11
Added support for proprietary devices in BACnet connector

apachler:
This feature makes it possible to inject proprietary device information the way BACpypes3 handels it.

To inject devices it is needed to import a custom class, atm this is only possible for advanced users since it is needed to implement a custom `AsyncBACnetConnector` class.

This PR makes sure the actual BACnet connector implementation is able to handle it.

### DIFF
--- a/thingsboard_gateway/connectors/bacnet/application.py
+++ b/thingsboard_gateway/connectors/bacnet/application.py
@@ -237,7 +237,7 @@ class Application(NormalApplication, ForeignApplication):
                 object_id = object['objectId']
                 if not isinstance(object_id, ObjectIdentifier):
                     obj_str = f"{object['objectType']},{object_id}"
-                    object_id = ObjectIdentifier(obj_str)
+                    object_id = vendor_info.object_identifier(obj_str)
 
                 object_class = vendor_info.get_object_class(object_id[0])
                 if object_class is None:
@@ -249,7 +249,7 @@ class Application(NormalApplication, ForeignApplication):
                     object['propertyId'] = {object['propertyId']}
 
                 for prop in object['propertyId']:
-                    property_identifier = PropertyIdentifier(prop)
+                    property_identifier = vendor_info.property_identifier(prop)
 
                     properties.append(property_identifier)
 


apachler:
I am working on a clean way to inject proprietary devices classes without the need of defining custom `AsyncBACnetConnect`.
Since i am not sure how long this will take, it would be nice to have that PR merged so everyone can use the more advanced way.

apachler:
Please do not merge this PR for now ...

apachler:
Feature done with dynamic loading of custom device extension.

In folder `extensions/bacnet/proprietary` there is a sample file how custom objects have to be defined. All files in this directory are imported dynamically on connector start.

apachler:
@imbeacon any comments?

samson0v:
@apachler, thanks for the PR. I will review it soon.

thingsboard_gateway/connectors/bacnet/bacnet_connector.py
```diff
@@ -73,6 +74,9 @@ def __init__(self, gateway, config, connector_type):
             self.__parse_ede_config()
             self.__log.debug('EDE config parsed')
 
+        # importing the proprietary package registers all available custom object types and properties
+        TBModuleLoader.import_package_files(connector_type, "proprietary")
```
samson0v: Perhaps it would be better to offer an option that allows the user to choose whether to load custom props or not.
To add something like a `loadProprietaryDevices` parameter in the connector configuration file:
```json
"application": {
    "objectName": "TB_gateway",
    "host": "0.0.0.0",
    "port": 47808,
    "loadProprietaryDevices": false,
    ....
  },
```

What do you think?

thingsboard_gateway/extensions/bacnet/proprietary/desigo_cc.py
```diff
@@ -0,0 +1,78 @@
+"""
```
samson0v: Does this mean that props and other stuff will be registered in bacpypes3 automatically on the fly, and these props will be available for reading?

samson0v (changes requested):

thingsboard_gateway/extensions/bacnet/proprietary/desigo_cc.py
apachler: yes, that is correct.

thingsboard_gateway/connectors/bacnet/bacnet_connector.py
apachler: Oh yes, that is correct, so implementing such an option makes sense then ;)

samson0v:
@apachler, also, please, revert changes for the uplink converter, as it was fixed.
We are planning a release tomorrow. Could you resolve the points above, so your feature will be included?

apachler:
> @apachler, also, please, revert changes for the uplink converter, as it was fixed. We are planning a release tomorrow. Could you resolve the points above, so your feature will be included?

done, hopefully i did it to your liking ;)

samson0v:
@apachler, well done! Thanks for your contribution!
I will merge your PR today.

PR #1884, opened 2025-07-15
Fixed data key filtering in BACnet connector uplink converter

apachler:
This PR fixes a problem with filtering of data point used for replacements in key expressions.

Before this change following config would lead to missing data points:
```json
{
  "devices": [
    {
      "attributes": [
        {
          "key": "description",
          "objectType": "device",
          "objectId": "2098183",
          "propertyId": "description"
        }
      ]
    }
  ]
}
```

After this change only correctly defined expression are respected to replace:
```json
{
  "devices": [
    {
      "attributes": [
        {
          "key": "${description}",
          "objectType": "device",
          "objectId": "2098183",
          "propertyId": "description"
        }
      ]
    }
  ]
}
```


samson0v:
@apachler, thanks for your time and efforts

PR #1998, opened 2025-10-27
Fixed reserved two-way RPC

Andry925:
Fixed mqtt reserved two way rpc


thingsboard_gateway/connectors/mqtt/mqtt_connector.py
```diff
@@ -961,7 +961,7 @@ def __process_rpc_request(self, content, rpc_config):
                                                             self.rpc_cancel_processing)
 
                 # Wait for RPC to be successfully enqueued, which never fails.
-                while self.__gateway.is_rpc_in_progress(expected_response_topic):
+                while not self.__gateway.is_rpc_in_progress(expected_response_topic):
```
Andry925: Removed "not" from this condition as it seems to be incorrect and may stuck here on many two-way RPC'S requests.

PR #2020, opened 2025-11-14
Fixed error with invalid logic using mask in bacnet application section

Andry925:
1. Fixed error with invalid logic using mask in bacnet application section.
2. Correct logic to validate netmask format.


samson0v:
@Andry925, duplicated logic
I will close this PR

PR #2066, opened 2026-02-10
Added overridable pre/post file processing hooks

cpavot:
### Summary

This pull request introduces two optional hook methods in the FTP connector lifecycle:

- `_on_file_preprocessing(...)`
- `_on_file_postprocessing(...)`

These hooks allow users to execute custom logic **before** and **after** a file is processed, without modifying the core FTP connector implementation. The default implementations are no-ops, so existing behavior remains unchanged.

### Motivation

In real-world FTP ingestion scenarios, users often need to perform deployment-specific actions around file processing, such as:

- custom logging or metrics
- moving or renaming processed files
- archiving workflows
- integration-specific housekeeping

Currently, implementing such logic requires patching the core connector. This PR provides an extensible alternative by exposing well-defined hook points.

### Design choices

- Hooks are implemented as `@staticmethod`s to avoid coupling with the connector internal state.
- All required context (`ftp`, `logger`, `file`) is explicitly passed as arguments.
- Default implementations do nothing, ensuring full backward compatibility.
- Hooks are invoked at precise points in the existing `_process_paths` workflow:
  - before file retrieval
  - after file processing and data submission

This keeps the change minimal, generic, and safe for upstream inclusion.

### Backward compatibility

- No configuration changes required
- No behavior changes unless hooks are explicitly overridden
- Existing FTP connectors continue to work as before

### Example usage (custom connector)

Below is an example of a **custom FTP connector** extending the base connector and overriding the post-processing hook to move processed files into a `processed/` directory.

This example is **not part of this PR** and is provided for illustration purposes only.

```python
import posixpath
from ftplib import error_perm, FTP
from logging import Logger

from thingsboard_gateway.connectors.ftp.ftp_connector import FTPConnector
from thingsboard_gateway.connectors.ftp.file import File

class FTPConnectorWithPostProcessing(FTPConnector):

    @staticmethod
    def _on_file_preprocessing(ftp : FTP, log : Logger, file : File):
        log.info("File %s : starting processing...", file.path_to_file)

    @staticmethod
    def _on_file_postprocessing(ftp : FTP, log : Logger, file : File):
        orig_dir = posixpath.dirname(file.path_to_file)
        filename = posixpath.basename(file.path_to_file)
        dest_dir = posixpath.join(orig_dir,"processed")

        try:
            ftp.mkd(dest_dir)
        except error_perm:
            # Already exists
            pass

        ftp.rename(file.path_to_file, posixpath.join(dest_dir, filename))
        log.info("File %s moved to %s", file.path_to_file, dest_dir)
````

### Scope of this PR

✔ Adds generic extension points
✔ Keeps core connector logic unchanged
✔ Enables clean customization via inheritance
✔ Does not introduce any deployment- or customer-specific behavior

### Notes

This change was motivated by production use cases and aims to reduce the need for maintaining long-lived forks of the FTP connector.

Feedback and suggestions are welcome !

### Additional fix: stable file sorting when using MDTM

This PR now also includes a small fix in the FTP file listing logic when `_with_sorting_files` is enabled.

Previously, files were indexed using the raw `MDTM` response as a dictionary key. This would lead to collisions when multiple files share the same modification timestamp (same second), causing some files to be silently overwritten and skipped.

The logic has been updated to:
- extract the MDTM timestamp explicitly
- use a collision-free composite key `(timestamp, full_path)`
- keep the exact same return structure and behavior

This ensures:
- no file loss when multiple files share the same MDTM value
- stable and deterministic sorting
- full backward compatibility for existing configurations

No functional behavior is changed unless file sorting is enabled.

### DIFF
--- a/thingsboard_gateway/connectors/ftp/path.py
+++ b/thingsboard_gateway/connectors/ftp/path.py
@@ -13,7 +13,8 @@
 #     limitations under the License.
 
 import os
-
+import posixpath
+from ftplib import error_perm
 from regex import compile
 
 from thingsboard_gateway.connectors.ftp.file import File
@@ -65,7 +66,8 @@ class Path:
 
             for folder_or_file in folder_and_files:
                 try:
-                    cur_file_name, cur_file_ext = folder_or_file.split('.')
+                    # rsplit() avoids issues with filenames containing multiple dots
+                    cur_file_name, cur_file_ext = folder_or_file.rsplit('.', 1)
                     if cur_file_ext in COMPATIBLE_FILE_EXTENSIONS \
                             and self.__is_file(ftp, folder_or_file):
                         if (file_name == file_ext == '*') \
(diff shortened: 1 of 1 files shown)


cpavot:
Hello @samson0v, here is a new PR ! There are a few arbitrary design choices, so don't hesitate if you have any feedback, I can update it. Thanks !

cpavot:
FYI, I just added another commit to fix a collision during file sorting.

samson0v:
Hi @cpavot, thanks for your contribution!
I will merge it.

A little advice:
We have a custom `TbLogger` class that inherit python `Logger`, so it is better to use it.

samson0v:
I will change the logger outside the scope of this PR.

cpavot:
Hi @samson0v, thanks for the feedback. Looking forward release of v3.8.3! ;)

PR #2068, opened 2026-02-11
pkg_resources deprecated in setuptools 82.0.0

cpavot:
## Context

Recent versions of `setuptools` (>= 82) removed `pkg_resources`.
Some ThingsBoard-related dependencies (e.g. `tb-mqtt-client` in `thingsboard-python-client-sdk`) still import `pkg_resources`, which causes the gateway to fail at startup with:

    ModuleNotFoundError: No module named 'pkg_resources'

## Changes

- Replaced `pkg_resources.get_distribution()` in   `thingsboard_gateway/tb_utility/tb_utility.py`  with `importlib.metadata` (Python standard library).
- Added a temporary `setuptools<82` constraint in the build configuration to preserve compatibility with external dependencies that still rely on `pkg_resources`.

## Impact

- No functional change in gateway behavior.
- Removes internal dependency on `pkg_resources`.
- Prevents runtime crash when building with recent `setuptools` versions.

The `setuptools<82` constraint can be removed once upstream dependencies migrate away from `pkg_resources`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -43,7 +43,7 @@ if [ ! -f "$FIRSTLAUNCH" ]; then\n\
 fi\n\
 echo "nameserver 8.8.8.8" >> /etc/resolv.conf\n\
 python /thingsboard_gateway/tb_gateway.py' > /start-gateway.sh && chmod +x /start-gateway.sh && \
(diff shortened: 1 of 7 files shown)


samson0v:
@cpavot Thanks for your contribution!
I will fix upstream dependencies.

cpavot:
Hi @samson0v, thanks ! :)

Issue #2077, opened 2026-02-17
[BUG] REST Connector send JSON as string


apachler:
**Connector name (If bug in the some connector):**
REST Connector

**Describe the bug**
I have updated the GW from v3.8.0 to v3.8.2 and now the REST Connector sends downlink JSON message as string to the endpoint.

I am using attribute update request mapping like:

``` json
  "requestsMapping": {
    "attributeRequests": [],
    "attributeUpdates": [
      {
        "HTTPMethod": "POST",
        "SSLVerify": false,
        "httpHeaders": {
          "Content-Type": "application/json"
        },
        "timeout": 3.0,
        "tries": 5,
        "allowRedirects": true,
        "deviceNameFilter": ".*",
        "attributeFilter": "AGILOXREST-occupation",
        "requestUrlExpression": "http://x.x.x.x",
        "valueExpression": "[{\"station\": \"${deviceName}\", \"command\": \"${attributeValue}\"}]"
      },
   ]
}
```

I have added some debug logs to find the problem...

``` log
2026-02-17 11:13:58.706 - |DEBUG| - [tb_gateway_service.py] - tb_gateway_service - _attribute_update_callback - 1862 - Attribute request received with content: "{'device': 'AH05-A01', 'data': {'AGILOXREST-occupation': 'occupy'}}"
2026-02-17 11:13:58.706 - |DEBUG| - [tb_gateway_service.py] - tb_gateway_service - _attribute_update_callback - 1863 - ()
2026-02-17 11:13:58.731 - |DEBUG| - [rest_connector.py] - rest_connector - on_attributes_update - 327 - Converted data: {'url': 'http://x.x.x.x', 'data': '[{"station": "AH05-A01", "command": "occupy"}]'}
2026-02-17 11:13:58.733 - |DEBUG| - [client.py] - client - _easy_log - 3264 - Sending PUBACK (Mid: 14)
2026-02-17 11:13:58.732 - |DEBUG| - [rest_connector.py] - rest_connector - __form_request_from_content_type - 524 - Forming request to URL: http://x.x.x.x
[...]

PR #2079, opened 2026-02-18
Fixed missing string conversion for downlink messages of type json

apachler:
REST Connector was missing json.loads if downlink message is not of type dict or list.


apachler:
This will fix https://github.com/thingsboard/thingsboard-gateway/issues/2077

apachler:
Already fixed by https://github.com/thingsboard/thingsboard-gateway/pull/2081

PR #2080, opened 2026-02-18
feat(poll_scheduler): add cron-based poll scheduling to all connectors

apachler:
# Add cron-based `pollSchedule` support for all connectors

## What

Adds an optional `pollSchedule` parameter that accepts cron expressions to control poll timing. This enables wall-clock-aligned polling (e.g. exactly at :00, :05, :10) and time-dependent schedules (e.g. faster polling during production, slower on weekends).

Also fixes a missing `pollPeriod` default in the Modbus Slave connector (`config['pollPeriod']` → `config.get('pollPeriod', 5000)`), which is the only connector that crashes when `pollPeriod` is omitted.

## Configuration

```json
// Simple — clock-aligned every 5 minutes
{ "pollSchedule": "*/5 * * * *" }

// Multiple schedules with labels
{ "pollSchedule": [
    { "cron": "*/1 6-22 * * 1-5", "label": "day-shift" },
    { "cron": "*/15 * * * 0,6",   "label": "weekend" }
  ]
}
```

```json
// Disabled (default in sample configs)
{ "pollSchedule": null }
```

`pollSchedule` takes precedence over `pollPeriod` when present. A value of `null` (or omitting the key entirely) keeps the original `pollPeriod` behavior unchanged.

> **Note:** All sample configs (`thingsboard_gateway/config/*.json`) now ship with `"pollSchedule": null` so the option is visible but inactive by default. To enable, replace `null` with a cron string or an array of schedule objects as shown above.

## Changes

- **New module**: `thingsboard_gateway/tb_utility/poll_scheduler.py` — `PollScheduler`, `PollScheduleEntry`, `compute_next_poll()`
- **New dependency**: `croniter>=1.3.8` (depends only on `python-dateutil`, already in the tree)
- **Bug fix**: `connectors/modbus/slave.py` — add missing `pollPeriod` default (5000ms)
- **All polling connectors**: Modbus, OPC-UA, BACnet, BLE, SNMP, CAN, FTP, ODBC, Request — minimal change: instantiate `PollScheduler` from config, use `compute_next_poll()` instead of `monotonic() + poll_period`

## Breaking changes

None.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,3 +20,4 @@ service-identity
 psutil
 cryptography
 PySocks
+croniter>=1.3.8
--- a/setup.py
+++ b/setup.py
@@ -88,6 +88,7 @@ setup(
         'service-identity',
         'psutil',
         'PySocks',
+        'croniter>=1.3.8',
     ],
     download_url='https://github.com/thingsboard/thingsboard-gateway/archive/%s.tar.gz' % version.VERSION,
     entry_points={
--- a/tests/unit/test_poll_scheduler.py
+++ b/tests/unit/test_poll_scheduler.py
@@ -1,0 +1,129 @@
+import pytest
+from time import monotonic
+from datetime import datetime
+from unittest.mock import patch
+
+from thingsboard_gateway.tb_utility.poll_scheduler import (
+    PollScheduleEntry, PollScheduler, compute_next_poll
+)
+
+
+# --- Config parsing ---
+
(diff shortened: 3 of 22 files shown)


samson0v:
Hi @apachler, thanks for your contribution!
Can we plan a review of this feature on the release v.3.8.4?

apachler:
Hi @samson0v, sure, let me know what you need from me ;)

PR #2085, opened 2026-02-23
Fixed escaping regex metacharacters for topics

### LINKED ISSUE #2084: [BUG] MQTT Connector Fails to Match Topics with Regex Metacharacters
**Connector name**
MQTT Connector

**Describe the bug**
The topic_to_regex() function in tb_utility.py does not escape regex metacharacters in topic filters before they are used with fullmatch() for internal message routing. This causes topics containing characters like parentheses (), brackets [], dots ., or other regex-special characters to silently fail matching. Messages are successfully received via MQTT subscription but are dropped during internal routing with a debug log entry: "Received message to topic ... with unknown interpreter data".

The root cause is in tb_utility.py line 114:
`topic.replace("+", "[^/]+").replace("#", ".+").replace('$', '\\$')`

This only converts MQTT wildcards (+, #) and escapes $, but does not escape other regex metacharacters. For example, a topic filter containing "(CCS2)" is interpreted as a regex capture group instead of literal parentheses, causing fullmatch() to fail.

Escaping the parentheses in the config (\\(CCS2\\)) is not a viable workaround because topicFilter is used for both MQTT subscription (mqtt_connector.py line 485, raw string) and regex matching (mqtt_connector.py line 628, converted string). Escaped parentheses would fix the regex but break the MQTT subscription since the broker receives literal backslashes.

**Steps to Reproduce**
1. Configure an MQTT connector with a topic filter containing parentheses:
```json
{
  "topicFilter": "my/topic/device name (model)/+",
  "converter": {
    "type": "json",
    "deviceInfo": {
      "deviceNameExpressionSource": "constant",
      "deviceNameExpression": "MyDevice",
      "deviceProfileExpressionSource": "constant",
      "deviceProfileExpression": "default"
    },
    "timeseries": [{ "type": "double", "key": "value", "value": "${value}" }]
  }
}
```
2. Publish an MQTT message to my/topic/device name (model)/data with payload {"value": 42}
3. Observe that the gateway log shows:
`Received message to topic "my/topic/device name (model)/data" with unknown interpreter data:`
4. The message is received but never routed to the converter — no device is created, no telemetry is stored

**Error traceback:**
No exception is raised. The message is silently dropped with a DEBUG level log:
```
|DEBUG| - [mqtt_connector.py] - mqtt_connector - _process_on_message - 695 - Received message to topic "my/topic/device name (model)/data" with unknown interpreter data:
```

**Versions**
OS: Debian (Docker container, ARM64)
ThingsBoard IoT Gateway version: 3.8.2
Python version: 3.11

**Additional context**
Affected source files:
- tb_utility.py line 114 (topic_to_regex — missing re.escape())
- mqtt_connector.py line 628 (fullmatch(regex, message.topic) — fails on unescaped metacharacters)
- mqtt_connector.py line 485 (self.__subscribe(mapping["topicFilter"], ...) — uses raw string, which is correct)
Any MQTT topic containing regex metacharacters ((, ), [, ], ., *, ?, {, }, |, ^) in the literal topic path will trigger this bug. The current workaround is to use + wildcards to skip topic levels containing these characters and extract the device name from the topic using deviceNameExpressionSource: "topic".

tijn-hedgehog:
## Summary

- Fix `topic_to_regex()` in `tb_utility.py` to escape all regex metacharacters using `re.escape()` before converting MQTT wildcards (`+`, `#`) to their regex equivalents
- Prevents `fullmatch()` from silently failing when topic filters contain characters like `(`, `)`, `.`, `[`, `]`, etc.
- Update `regex_to_topic()` accordingly since `re.escape()` now handles `$` escaping

Fixes #2084

### DIFF
--- a/thingsboard_gateway/tb_utility/tb_utility.py
+++ b/thingsboard_gateway/tb_utility/tb_utility.py
@@ -17,7 +17,7 @@ from getpass import getuser
 from logging import getLogger, setLoggerClass
 from os import environ
 from platform import system as platform_system
-from re import search, findall
+from re import escape as re_escape, search, findall
 from time import monotonic, sleep
 from typing import Union, TYPE_CHECKING
 from uuid import uuid4
@@ -111,11 +111,11 @@ class TBUtility:
 
     @staticmethod
     def topic_to_regex(topic):
-        return topic.replace("+", "[^/]+").replace("#", ".+").replace('$', '\\$')
+        return re_escape(topic).replace(r'\+', '[^/]+').replace(r'\#', '.+')
 
     @staticmethod
     def regex_to_topic(regex):
-        return regex.replace("[^/]+", "+").replace(".+", "#").replace('\\$', '$')
+        return regex.replace('[^/]+', '+').replace('.+', '#')
 
     @staticmethod
     def get_value(expression, body=None, value_type="string", get_tag=False, expression_instead_none=False):


samson0v:
Hi @tijn-hedgehog, thanks for your contribution!
I will merge your PR.

PR #2087, opened 2026-02-23
Correct rate limiting configuration key names

apachler:
Fix incorrect configuration key names used for telemetry and device rate limiting settings. The configuration loader expects simplified key names (rateLimits, dpRateLimits, deviceRateLimits, deviceDpRateLimits) but the code was checking for more verbose names with redundant prefixes. This corrects the mapping between configuration keys and internal rate limit settings.


samson0v:
@apachler Many thanks for your contribution!

PR #2096, opened 2026-02-26
default port for bacnet; ensure INT for port and mask; fixing default BACnet config

majuss:
default BACnet config:
- ensure mask and port are integer
- application IP cannot be 0.0.0.0 when mask is /24, set a plausible IP

bacnet_connector:
- ensure Port and Mask are read as Integer, everything else will throw errors (string doesnt make any sense)
- ensure default BACnet port even when parameter port is completely skipped

bacnet_tests:
- adjusted tests to integer

### DIFF
--- a/thingsboard_gateway/connectors/bacnet/bacnet_connector.py
+++ b/thingsboard_gateway/connectors/bacnet/bacnet_connector.py
@@ -216,7 +216,7 @@ class AsyncBACnetConnector(Thread, Connector):
 
     async def __start(self):
         if not self.__is_valid_application_device_section():
-            self.__log.error('Can not start connector due to invalid application section in config.')
+            self.__log.error('Can not start connector due to invalid application/devices section in config.')
             return
 
         if self.__config.get('foreignDevice', {}).get('address', ''):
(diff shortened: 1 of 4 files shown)


majuss:
@samson0v can you please have a look. Thank you!

majuss:
@samson0v push

thingsboard_gateway/connectors/bacnet/bacnet_connector.py
```diff
@@ -239,6 +239,28 @@ async def __start(self):
                              self.indication_callback(),
                              self.__application.confirmation_handler())
 
+    def __validate_and_normalize_port(self, section: dict, section_name: str) -> bool:
+        port = section.get('port')
+        if port is None:
+            section['port'] = 47808
+            return True
```
samson0v, line 246: We need to cast the port to an integer because the default remote config uses device port "47808" (as a string), so the device will be permanently invalid from startup.

We will change the remote default config, but the connector should pass the port as a string for now.

samson0v (changes requested):

thingsboard_gateway/connectors/bacnet/bacnet_connector.py
majuss: This PR also changes the default config to INT

PR #2111, opened 2026-03-24
Added support for offline Docker builds and image exports for air-gapped environments

wbg1991:
This PR introduces support for building and exporting Docker images designed specifically for air-gapped (closed-network) environments.

While the repository currently provides .deb installers for offline deployment, there is a growing demand for containerized deployments even in restricted environments for better orchestration and maintenance. This update bridge that gap by providing a native Docker-based offline workflow.

Changes:
New Dockerfile.offline: Uses requirements-full.txt instead of the standard requirements to pre-install all connector dependencies during the build stage.

.dockerignore Update: Modified to include requirements-full.txt in the build context so it can be accessed by the offline Dockerfile.

--offline Flag: Added to the build script to trigger the build using the offline-ready Dockerfile.

--save Flag: Added to allow users to export the built image directly into a .tar.gz archive, making it easy to transfer to servers without internet access.

Why this is needed:
In many enterprise or high-security production environments, direct internet access is prohibited. While .deb packages work, many modern infrastructures prefer Docker for consistency. These changes allow users to build a "heavy" image once in a networked environment and deploy it seamlessly in a fully offline environment.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,7 @@
 !README.md
 !setup.py
 !requirements.txt
+!requirements-full.txt
 
 # ignore on every level
 **/__pycache__/
--- a/docker/Dockerfile.offline
+++ b/docker/Dockerfile.offline
@@ -1,0 +1,63 @@
+FROM python:3.11-slim AS build
+
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+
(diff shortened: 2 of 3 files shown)


samson0v:
Hi @wbg1991, thanks for your contribution!

PR #2117, opened 2026-04-01
Prevent custom connectors from colliding

ecoutier:
## Context

This PR is linked to the feature request I mentioned in the [discord server](https://discord.com/channels/1458396495610122526/1487081016769380483/1487081016769380483).
When using remote config with multiple custom connectors of the same type, only the last one was ever instantiated. This is because `_implemented_connectors` was keyed by `connector_type`, so each new connector of the same type would silently overwrite the previous entry:
```python
self._implemented_connectors[connector_type] = connector_class
```
I may not have the full context on why it was originally designed this way, happy to get feedback or alternative approaches if there's a reason I'm missing.

## Changes

`_implemented_connectors` is now keyed by connector name instead of connector type, both at registration (`_load_connectors`) and at lookup (`__connect_with_connectors`, `__init_and_start_regular_connector`). Since names are supposedly unique per connector instance, multiple custom connectors of the same type can now coexist.

## Impact

This is a workaround pending proper multi-custom-connector support in the UI. 
This has worked on my part to be able to have 2 different custom connectors at the same time, with 2 different classes. 
Please note that RPC calls have not been tested
Only covers the remote config scenario : for non remote config, it is still possible to have a folder per custom connector.

### DIFF
--- a/thingsboard_gateway/gateway/tb_gateway_service.py
+++ b/thingsboard_gateway/gateway/tb_gateway_service.py
@@ -910,7 +910,7 @@ class TBGatewayService:
                                       connector_config_from_main['name'])
                             continue
                         else:
-                            self._implemented_connectors[connector_type] = connector_class
+                            self._implemented_connectors[connector_config_from_main['name']] = connector_class
                     elif connector_type == "grpc":
                         if connector_config_from_main.get('key') == "auto":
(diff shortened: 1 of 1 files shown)


samson0v:
Hi @ecoutier, thanks for your contribution!
I will merge your PR.

PR #2120, opened 2026-04-03
fix: preserve original telemetry timestamp on periodic resend

### LINKED ISSUE #2119: Report strategy overwrites original telemetry timestamp with current time on periodic resend
## Description

When using `ON_REPORT_PERIOD` or `ON_CHANGE_OR_REPORT_PERIOD` report strategies, the periodic reporting thread overwrites the original telemetry timestamp with `int(time() * 1000)` (current wall-clock time) before re-sending. This makes it impossible for downstream consumers to detect stale data. The timestamp always looks fresh even if the source device stopped updating hours ago.

## Where

`thingsboard_gateway/gateway/report_strategy_service.py`, in `__periodical_reporting()`:

```python
if report_strategy_data_record.is_telemetry():
    # data_entry.add_to_telemetry(TelemetryEntry({key: value}, report_strategy_data_record.get_ts())) # Can be used to keep first ts, instead of overwriting it with current ts # noqa
    current_ts = int(time() * 1000)
    data_entry.add_to_telemetry(TelemetryEntry({key: value}, current_ts))
    report_strategy_data_record.update_ts(current_ts)
```

The commented-out line preserves the original timestamp, but the active code replaces it.

## Impact

With `ON_CHANGE_OR_REPORT_PERIOD`, if an OPC-UA node or MQTT device goes stale (stops updating its value), the gateway keeps re-sending the last cached value with a **fresh timestamp** every report period. ThingsBoard shows the telemetry as recently updated, so there's no way to tell the device is actually offline or stuck.

This breaks any staleness detection on the ThingsBoard side (dashboards, rule chains, frontends checking `ts` age).

## Expected behavior

Periodic resends should preserve the original telemetry timestamp from when the value was last actually received from the device. The `receivedTs` in metadata already tracks when the gateway forwarded it, so that covers the "last forwarded" use case.

## Suggested fix

```python
if report_strategy_data_record.is_telemetry():
    data_entry.add_to_telemetry(TelemetryEntry({key: value}, report_strategy_data_record.get_ts()))
```

Remove the `current_ts` override and the `update_ts(current_ts)` call. This way:
- Periodic resends keep the original source timestamp → staleness is detectable
- Value changes still get a fresh timestamp via `filter_datapoint_and_cache`
- `receivedTs` in metadata still reflects when the gateway last forwarded

## Reproduction

1. Configure a device with `ON_CHANGE_OR_REPORT_PERIOD` strategy (e.g. 60s report period)
2. Send one telemetry value, then stop the device
3. Observe ThingsBoard latest telemetry — the `ts` keeps updating every 60s even though the device is offline

## Version

Latest `main` branch (also present in 3.x releases)

tijn-hedgehog:
## Summary

- Fixes periodic reporting overwriting the original telemetry timestamp with current wall-clock time in `__periodical_reporting()`
- Periodic resends now preserve the original source timestamp (`report_strategy_data_record.get_ts()`) so downstream consumers can detect stale data
- `receivedTs` in metadata still reflects when the gateway forwarded the data

## Problem

When using `ON_REPORT_PERIOD` or `ON_CHANGE_OR_REPORT_PERIOD` strategies, the periodic reporting thread replaced the original telemetry timestamp with `int(time() * 1000)` before re-sending. This made stale data appear fresh in ThingsBoard. If a device stopped updating, the `ts` kept advancing every report period, breaking any staleness detection in dashboards, rule chains, or frontends.

## Change

In `report_strategy_service.py`, `__periodical_reporting()`:

```python
# Before (overwrites timestamp):
current_ts = int(time() * 1000)
data_entry.add_to_telemetry(TelemetryEntry({key: value}, current_ts))
report_strategy_data_record.update_ts(current_ts)

# After (preserves original timestamp):
data_entry.add_to_telemetry(TelemetryEntry({key: value}, report_strategy_data_record.get_ts()))
```

The commented-out line that already existed for this purpose has been activated and the timestamp-overwriting code removed.

## Test plan

- [ ] Configure a device with `ON_CHANGE_OR_REPORT_PERIOD` strategy (e.g. 60s report period)
- [ ] Send one telemetry value, then stop the device
- [ ] Verify ThingsBoard latest telemetry `ts` stays at the original value (does not advance every 60s)
- [ ] Verify `receivedTs` in metadata still updates on each resend
- [ ] Verify that when a new value arrives, `ts` updates correctly via `filter_datapoint_and_cache`
- [ ] Test with `ON_REPORT_PERIOD` strategy as well

Fixes #2119

### DIFF
--- a/thingsboard_gateway/gateway/report_strategy/report_strategy_service.py
+++ b/thingsboard_gateway/gateway/report_strategy/report_strategy_service.py
@@ -233,10 +233,7 @@ class ReportStrategyService:
 
                     data_entry = data_to_report[data_report_key]
                     if report_strategy_data_record.is_telemetry():
-                        # data_entry.add_to_telemetry(TelemetryEntry({key: value}, report_strategy_data_record.get_ts())) # Can be used to keep first ts, instead of overwriting it with current ts # noqa
-                        current_ts = int(time() * 1000)
-                        data_entry.add_to_telemetry(TelemetryEntry({key: value}, current_ts))
-                        report_strategy_data_record.update_ts(current_ts)
+                        data_entry.add_to_telemetry(TelemetryEntry({key: value}, report_strategy_data_record.get_ts()))
 
                         reported_data_length += 1
                     else:


iamkaran:
Correct me if i am wrong but doesn't this defeat the purpose of `ON_REPORT_PERIOD` as it serves the purpose of maintaining connection activity by sending data periodically even if it is stale.
Its also mentioned in the [docs](https://thingsboard.io/docs/iot-gateway/features-overview/report-strategy/):
image

The `current_ts` behavior is intentional: the periodic updates are heartbeats disguised as telemetry

tijn-hedgehog:
If that's the case, then I propose a new reporting strategy that either sends data on change, or every reporting period (but preserves the timestamp so the user knows if it's stale). 

In my customer dashboard, I want to show  data updates periodically (eg, every 30s), but also instant updates when values have changed. When a device has gone stale, the current reporting strategy would keep sending the stale data as if it were fresh, misleading the customer. Currently, there is no (to my understanding, correct me if I'm wrong) report strategy that has this behavior.

iamkaran:
can we close this PR then? :octocat:

Stefan-1313:
I hope I'm not too late. This is exaclty the feature I'm also looking for.
I'd like to add a concrete use case in support of this PR's direction.

I use `ON_CHANGE_OR_REPORT_PERIOD` to avoid storing excessive telemetry while still keeping time-series queries simple — having at least one real data point per reporting interval means I can query a bounded time range and always get a meaningful result, without expensive *"look back until last known value"* queries.

The current behavior breaks this: when a device goes silent, `ON_CHANGE_OR_REPORT_PERIOD` keeps forwarding the last known value with a fresh timestamp, so **the absence of new data is invisible**. *No-data is also data* — it tells me a device may be offline, a sensor may have failed, or a process has stopped. Hiding that with synthetic heartbeat timestamps is genuinely misleading in dashboards and rule chains.

My preferred behavior:
- If at least one real measurement was received during the reporting interval → send it, preserving the original source timestamp.
- If **no** measurement was received → send nothing. Let the gap speak for itself.

If I understand correctly, this is exactly what this PR moves toward. The concern that `ON_REPORT_PERIOD` serves as a "heartbeat" is valid for some users, but it shouldn't be the only mode. I'd suggest making it configurable with a flag, e.g. `suppress_on_no_data: true/false` (defaulting to `false` for backward compatibility), so both behaviors are supported:

- `suppress_on_no_data: false` (current/default): keeps the heartbeat behavior. Ideally preserving the original `last datapoint timestamp` so staleness detection is still possible but this part is optional but this is what this pull request does if I understand correctly.
- `suppress_on_no_data: true`: no real data in the interval = no telemetry written. Clean gaps in the time series.

This makes the feature backward compatible while giving users like me — and presumably the PR author's customers — a way to build honest dashboards.

Happy to help draft or test this if the maintainers are open to it.

samson0v:
Hi @tijn-hedgehog and @Stefan-1313, thanks for your interest in ThingsBoard IoT Gateway!
Can you explain to me why you can't use `ON_RECEIVED` report strategy (https://thingsboard.io/docs/iot-gateway/features/report-strategy/#strategy-types)?

Stefan-1313:
@samson0v 

Thanks for taking a look @samson0v — apologies if I'm missing something.

ON_RECEIVED does give honest timestamps: every stored point is a real measurement, and a silent device leaves a real gap. The reason it doesn't fit is volume — it writes every poll cycle, so a value that sits still for hours still produces a point per poll. We could poll slower, but that also delays how quickly we see a change.

What we're after, per report period:
1. value changed → written immediately
2. fresh data received but value identical → one point, carrying the real timestamp of that measurement
3. nothing received from the device → nothing written

As I mentioned earlier in this thread, the report period is what makes our time-series queries cheap: a guaranteed point per interval means we can query a bounded window and always get a meaningful result, instead of scanning backwards until the last known value. That guarantee is only worth anything if the point is a real measurement, which is exactly why 2 and 3 have to be distinguishable.

And that distinction is the part I think is currently missing. A repeated cached value isn't evidence that the value stayed the same — only a fresh measurement can tell you that. Today the periodic branch treats "received and identical" and "received nothing at all" the same way: it resends the cache with wall-clock time, so an offline device is indistinguishable from a steady one.

As far as I can tell, ON_RECEIVED and ON_CHANGE both cover 1 and 3 but not 2, from opposite directions — ON_RECEIVED writes every identical sample, ON_CHANGE writes none of them. ON_CHANGE_OR_REPORT_PERIOD covers 1 and 2, but not 3.

One related question while I'm here: is there any way to set a minimum interval between change-driven points? As I read it, ON_CHANGE is writing every change with no throttle, so a fast-changing value produces a point per change. A minimum spacing (for example 1 sec) would bound the write rate from the other side. So ON_REPORT creates a maximum interval between writes (if there is real data) and the ON_CHANGE to have a (configurable) minimum interval between writes to avoid fast flapping points to overload the gateway.  Though that's a separate concern from the timestamp issue above.

If there's a way to get 1–3 with the strategies as they are today, I'd like to hear it.

PR #2137, opened 2026-06-03
Fixed device duplicating when the device has both ID and IP parameters in the config

majuss:
This else statement is creating duplicate local device setups when IP and Bacnet ID are inside the config json


samson0v:
Hi @majuss, thanks for your contribution!

PR #2151, opened 2026-07-27
Fix coil decoding order

### LINKED ISSUE #2146: decode coils return incorrect result
https://github.com/thingsboard/thingsboard-gateway/blob/bbede93b64c676e9e6684f5b29a97ab41b99ca96/thingsboard_gateway/connectors/modbus/bytes_modbus_uplink_converter.py#L284

for coils decoded data, this line of code will fetch the trailing items rather than the leading items. It results in wrong data.

I tried following code to verify the logic:

```
from pymodbus.constants import Endian
from pymodbus.payload import BinaryPayloadDecoder

if __name__ == '__main__':
    objects_count = 2

    byte_order = Endian.LITTLE
    word_order = Endian.BIG

    # mock coils data
    encoded_data = [True, False, False, False, False, False, False, False]

    decoder = BinaryPayloadDecoder.fromCoils(encoded_data, byteorder=byte_order, _wordorder=word_order)
    decoded = decoder.decode_bits()
    decoded_lastbyte = decoder.decode_bits()
    decoded += decoded_lastbyte
    decoded = decoded[len(decoded) - objects_count:]
    print('result: ', decoded)  # it prints [False, True], not [True, False]
```

after investigating, this issue is caused by the reverse of bits in the BinaryPayloadDecoder.fromCoils(). 

Image

By design, the original bits in ReadCoilsResponse is in the same order as the actual coils. when creating decoder, the code is supposed to pack the bits to bytes in the reversed order and at last at decoding stage, read the bits back reversely again. so first right order(response bits) -> reverse order(decoder packing) -> reversed reverse order = right order(decode) will return correct result. However, in the BinaryPayloadDecoder.fromCoils() method, it reverses the original coils bits prior to packing it into bytes. So the processing sequence become: original bits(coils bits) -> reversed bits(from coils) -> reversed reversed bits = original bits(packing) -> reversed bits(decode). And then, the code fetches the bits from beginning, resulting incorrectly bit value:

        if lower_type in ['bit', 'bits']:
            decoded = decoder_functions[lower_type]()
            decoded_lastbyte = decoder_functions[lower_type]()
            decoded += decoded_lastbyte
            decoded = decoded[len(decoded)-objects_count:]

Jerry2661:
## SUMMARY
- preserve the leading coil bits when limiting decoded values to 'objectsCount'
- add a regression test for the reported bit-order case

Fixes #2146

## TESTING
- added 'test_decode_coils_preserves_leading_bit_order'
- not run locally; changes were made through the GitHub web editor

### DIFF
--- a/tests/unit/converters/modbus/test_bytes_modbus_uplink_converter.py
+++ b/tests/unit/converters/modbus/test_bytes_modbus_uplink_converter.py
@@ -214,6 +214,22 @@ class ModbusConverterTests(BaseUnitTest):
         result = converter.convert(test_modbus_convert_config, test_modbus_body_to_convert)
         self.assertDictEqual(result, test_modbus_result)
 
+    def test_decode_coils_preserves_leading_bit_order(self):
+        converter = object.__new__(BytesModbusUplinkConverter)
+        config = {
+            "functionCode": 1,
+            "type": "bits",
+            "objectsCount": 2,
+        }
+        encoded_data = [True, False, False, False, False, False, False, False]
+        result = converter.decode_data(
+            encoded_data,
+            config,
+            Endian.LITTLE,
+            Endian.BIG,
+        )
+        self.assertEqual([True, False], result)
+
 
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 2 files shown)


samson0v:
Hi @Jerry2661, thanks for your contribution!
Unfortunately, the test failed:
```
Expected :[True, False]
Actual   :[False, False]
```

Please fix your PR, and I will merge it.

PR #2152, opened 2026-07-28
Unpin Python "packaging"

mschwan-phytec:
Do not pin the Python library "packaging" to an old version anymore. Newer versions do not seem to contain any incompatibilities.

Pinning also blocks thingsboard-gateway from using the system-wide installed packaging library. This is relevant for e.g. embedded platforms, natively installing thingsboard-gateway and its dependencies.

This fixes the following error message during startup:
```
...
Mar 13 15:35:24 mydevice python3[695]: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
Mar 13 15:35:24 mydevice python3[695]: thingsboard-gateway 3.8.3 requires packaging==23.1, but you have packaging 26.0 which is incompatible.
...
```

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,5 +1,5 @@
 setuptools
-packaging==23.1
+packaging
 jsonpath-rw
 regex
 pip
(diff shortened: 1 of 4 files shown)


samson0v:
Hi @mschwan-phytec, thanks for your contribution!

PR #2168, opened 2026-08-30
Fixed partial records handling in file storage

### LINKED ISSUE #2167: [BUG] File storage can consume partially written records
**Connector name (If bug in the some connector):**
Not connector-specific. This affects the gateway file event storage.

**Describe the bug**
`EventStorageReader` treats any non-empty result from `readline()` as a complete Base64 record. When the reader reaches a file while `EventStorageWriter` is still appending that record, `readline()` can return the bytes currently available at EOF without a line terminator.

The partial Base64 payload may decode successfully but produce truncated JSON. The invalid message can then remain in the current batch and be returned repeatedly because it is never acknowledged as successfully published.

There is a second data-loss path during recovery: when restoring a non-zero position from `state_file`, `get_or_init_buffered_reader()` calls `readline()` once after the requested number of lines has already been skipped. This consumes the first pending record. If that record was the partial line preserved by a newline guard, a gateway restart can still skip its prefix and later expose only its suffix.

**Steps to Reproduce**
1. Create a file-storage data file containing one complete Base64-encoded JSON record followed by the first part of a second record, without `\n`.
2. Read the file with `EventStorageReader`.
3. Observe that the current implementation returns the incomplete second payload instead of waiting for the terminating newline.
4. For the restart variant, acknowledge the first complete record, recreate `EventStorageReader` from the saved position, append the rest of the second record plus `\n`, and read again.
5. Observe that the restored reader has consumed one extra line and returns only a suffix of the second JSON message.

**Expected behavior**
The reader should only consume newline-terminated records, rewind to the beginning of a partial line, preserve ordering across files, and restore the exact first unacknowledged record after restart.

**Error traceback (If available):**
The downstream symptom is a repeating JSON decode failure for the same stored message. The exact traceback depends on where the malformed payload is deserialized.

- OS: reproduced on Linux; the reader behavior is platform-independent
- ThingsBoard IoT Gateway version: 3.8.4 and current `master` (`726ed5c4`)
- Python version: reproduced on Python 3.11; deterministic regression tests also exercise 3.10-3.13

**Additional context**
A deterministic unit-test suite is ready covering:
- a partial first record;
- a complete record followed by a partial record;
- a partial current file when a next file already exists;
- restoring the reader after acknowledging a preceding record.

gustavovuolo:
Fixes #2167

## Problem

File storage writes each message as a Base64-encoded JSON record terminated by a newline. The writer and reader operate concurrently, but `EventStorageReader` accepted every non-empty result from `readline()` as a complete record.

At EOF, `readline()` may return the bytes currently available while the writer is still appending the record. A partial Base64 payload can decode successfully into truncated JSON. That invalid message may then remain in the current batch and be retried indefinitely because it is never acknowledged as published.

A related recovery bug made a newline guard insufficient on its own: restoring a non-zero state position called `readline()` once more after all acknowledged lines had already been skipped. After a restart, that consumed the first pending record (or its currently available prefix).

## Changes

- Record the byte offset before each `readline()` call.
- Only consume records terminated by `\n` (`\r\n` remains accepted).
- Seek back to the saved offset and return the completed part of the batch when the current line is still being written.
- Stop before file-transition logic so a partial current file is not removed when a newer file exists.
- Restore the state pointer by reading exactly the acknowledged number of lines, without consuming the next record.
- Add deterministic unit coverage for partial reads, ordering, file transitions, and restart recovery.

No configuration format or public API changes.

## Verification

The regression tests fail against the previous reader behavior and pass with this change.

```text
python -m unittest tests.unit.service.test_file_event_storage_reader -v
Ran 4 tests ... OK
```

The same four tests passed on Python 3.10, 3.11, 3.12, and 3.13. I also ran:

- the existing file-storage test and adjacent timestamp resolver tests (9 tests, all passing);
- `compileall` for the changed storage package and test;
- `pip check` and `git diff --check`;
- a build using the repository's official `docker/Dockerfile`.

In an isolated Docker deployment based on Gateway 3.8.4, the guarded reader processed hundreds of complete stored records with no JSON decode errors, storage-processing errors, duplicates, or gateway restarts.

### DIFF
--- a/tests/unit/service/test_file_event_storage_reader.py
+++ b/tests/unit/service/test_file_event_storage_reader.py
@@ -1,0 +1,138 @@
+#     Copyright 2026. ThingsBoard
+#
+#     Licensed under the Apache License, Version 2.0 (the "License");
+#     you may not use this file except in compliance with the License.
+#     You may obtain a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#     Unless required by applicable law or agreed to in writing, software
+#     distributed under the License is distributed on an "AS IS" BASIS,
+#     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#     See the License for the specific language governing permissions and
+#     limitations under the License.
+
+from base64 import b64encode
+from json import dump, dumps
+from logging import getLogger
+from os import sep
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from unittest import TestCase
+
+from thingsboard_gateway.storage.file.event_storage_files import EventStorageFiles
+from thingsboard_gateway.storage.file.event_storage_reader import EventStorageReader
+from thingsboard_gateway.storage.file.file_event_storage_settings import FileEventStorageSettings
+
+
+LOG = getLogger("TEST")
+LOG.trace = LOG.debug
+
+
+class TestFileEventStorageReader(TestCase):
+    DATA_FILE = "data_test.txt"
+    STATE_FILE = "state_file.txt"
+
+    def setUp(self):
+        self.temporary_directory = TemporaryDirectory()
+        self.data_directory = Path(self.temporary_directory.name)
+        self.data_file = self.data_directory / self.DATA_FILE
+        self.state_file = self.data_directory / self.STATE_FILE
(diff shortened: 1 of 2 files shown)


samson0v:
Hi @gustavovuolo, thanks for your contribution!

samson0v (approved):

PR #2176, opened 2026-09-07
Fixed connector log intervals and repeated size rollover

Cybergrany:
Connector and converter file handlers multiply the configured rotation interval twice when cloned. A three-day template becomes a roughly 710-year interval, **creating a plausible disk-exhaustion risk, particularily if using custom connectors with high log density.** Separately, repeated size rollovers can select an existing timed archive name, causing newer Python handlers to stop rotating or older handlers to overwrite retained records.

There's two production paths that excercise the bug:
- Creating connector/converter loggers. init_logger() copies the handler attached to the configured connector or converter logger, giving the copy its own filename. For example, the [Modbus connector’s](https://github.com/thingsboard/thingsboard-gateway/blob/d4529d1a26bb0739a862d679d4d15f3b2056a8ca/thingsboard_gateway/connectors/modbus/modbus_connector.py#L84) constructor initializes both loggers this way.
- Refreshing logging configuration. [update_file_handlers()](https://github.com/thingsboard/thingsboard-gateway/blob/d4529d1a26bb0739a862d679d4d15f3b2056a8ca/thingsboard_gateway/tb_utility/tb_logger.py#L251) creates replacement copies for existing connector/converter loggers. The gateway invokes this during startup and when remote logging configuration changes. 

The two commits address this:

- Preserve the constructor's interval units when copying connector and converter handlers.
- Disambiguate reused archive timestamps with a numeric suffix and prune archives in timestamp/numeric-index order. Determine the next index from existing archives so pruning and restarts cannot cause a newer archive to reuse an older index.

Validation:

```bash
python3 -B -m unittest tests.unit.service.test_tb_rotating_file_handler -v
python3 setup.py build_py
```

Added nine tests covering both factories across schedule units and non-default intervals; repeated same-second and advancing-clock midnight rollovers; exact retained records; backupCount values 0, 1, and 4; delayed opening; restarts; existing archives; custom string/path namers; subsequent timed rotation; and spring/autumn daylight-saving transitions.


thingsboard_gateway/tb_utility/tb_rotating_file_handler.py
```diff
@@ -60,6 +66,34 @@ def shouldRolloverOnSize(self, record):
...
+                    break
+
+    def getFilesToDelete(self):
+        if self.maxBytes <= 0:
+            return super().getFilesToDelete()
+        if self.backupCount <= 0:
+            return []
+        # Sort the collision index numerically: .10 is newer than .9.
+        archives = sorted(self._get_archive_files())
+        return [filename for _, _, filename in archives[:-self.backupCount]]
```
samson0v, line 95: What is the purpose of this method?
Also, it seems that it is never used.

samson0v (changes requested):

thingsboard_gateway/tb_utility/tb_rotating_file_handler.py
samson0v: @Cybergrany

Cybergrany: This overrides [TimedRotatingFileHandler.getFilesToDelete()](https://github.com/python/cpython/blob/v3.11.15/Lib/logging/handlers.py#L373) 
The inherited doRollover() calls self.getFilesToDelete() when backupCount > 0, so it's called from Python’s standard library.
My collision fix introduces numbered archives sharing a timestamp. This override sorts those suffixes numerically, ensuring .10 is considered newer than .9, and selects the oldest archives for deletion according to backupCount. The inherited doRollover() performs the deletion.
When size rotation is disabled, it delegates to the original implementation.

samson0v:
@Cybergrany, thanks for your contribution!

PR #2183, opened 2026-09-15
Implemented shared attribute sync

Prus-Oleksandr:
Add shared attributes synchronization for S7 connector upon reconnection


thingsboard_gateway/connectors/s7/s7_connector.py
```diff
@@ -211,6 +211,11 @@ async def _read_data_from_devices(self) -> None:
                     continue
 
                 results = await device.read_configured_data()
+
+                if not device.is_connected():
+                    self.__delete_device_from_platform(device)
+                    continue
```
samson0v: See snap7.client.Client class

samson0v (changes requested):

thingsboard_gateway/connectors/s7/entities/device.py
```diff
@@ -111,6 +115,12 @@ def stop(self) -> None:
             self._log.error(
                 f"Failed to disconnect from Logo device '{self.config.device_name}' at {self.config.address}:{self.config.port}: {e}")  # noqa: E501
 
+    def on_disconnect(self) -> None:
+        if self.config.device_name in self._gateway.get_devices(connector_id=self._connector_id):
+            self._log.warning('Device %s is disconnected, removing it from the platform.',
+                              self.config.device_name)
+            self._gateway.del_device(self.config.device_name)
```
samson0v: Maybe it is better to pass a callback from the connector instance to avoid passing the gateway and connector id

samson0v (changes requested):

thingsboard_gateway/connectors/s7/entities/device.py
```diff
@@ -67,18 +69,18 @@ def _load_converter(self, converter_type: str, converter_logger):
             self._log.exception(
                 'Failed to load %s converter for %s device: %s',
                 converter_type, self.config.device_name, e)
-
     @staticmethod
-    def create_device_from_config(logger, converter_logger, config: dict, reading_request_queue) -> 'Device':
+    def create_device_from_config(logger, converter_logger, config: dict, reading_request_queue,
+                                   on_disconnect=None) -> 'Device':
```
samson0v: Do we really need to make `on_disconnect` an optional argument?

samson0v (changes requested):

thingsboard_gateway/connectors/s7/entities/device.py
```diff
@@ -111,6 +114,10 @@ def stop(self) -> None:
             self._log.error(
                 f"Failed to disconnect from Logo device '{self.config.device_name}' at {self.config.address}:{self.config.port}: {e}")  # noqa: E501
 
+    def on_disconnect(self) -> None:
+        if self._on_disconnect_callback is not None:
```
samson0v: `self._on_disconnect_callback` is required

samson0v (changes requested):

